PR #8, opened 2022-12-08
fix: fqcn in module params

### LINKED ISSUE #7: Search / replace issue
Hi! I was looking into this collection, coming from https://github.com/cloudalchemy/ansible-node-exporter/issues/266#issuecomment-1327202944 and decided to try it out.

I noticed that for example here: https://github.com/prometheus-community/ansible/blob/main/roles/node_exporter/tasks/configure.yml#L7 there seems to be a search/replace issue with the group parameter (presumably to use fully quantified roles everywhere?)

I am not sure where else this occurs as I haven't searched outside of this file.

Thanks!

gardar:
Fixes fqcn's that had slipped in as module parameters and other minor lint fixes.

fixes #7

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,46 +1,34 @@
 ---
-### REQUIRED
-# The namespace of the collection. This can be a company/brand/organization or product namespace under which all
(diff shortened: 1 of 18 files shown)


SuperQ (approved):
Thanks!

PR #10, opened 2022-12-23
node_exporter: Change command parameters for ansible 2.14 compatibility

### LINKED ISSUE #9: node_exporter: Role doesn't work with Ansible 7.0.0
Due to a breaking change in [Ansible 2.14](https://github.com/ansible/ansible/blob/stable-2.11/changelogs/CHANGELOG-v2.11.rst#deprecated-features), using the `warn` parameter on `command` triggers an error: 

```
TASK [cloudalchemy.node_exporter : Gather currently installed node_exporter version (if any)] **********************************************************************************************
fatal: [node-1]: FAILED! => changed=false
  msg: 'Unsupported parameters for (ansible.legacy.command) module: warn. Supported parameters include: _raw_params, executable, strip_empty_ends, stdin_add_newline, creates, chdir, removes, stdin, _uses_shell, argv.'
```

(This issue is a copy of https://github.com/cloudalchemy/ansible-node-exporter/issues/276)

cudevmaxwell:
Use `ignore_errors` instead of warn parmeter. Fixes #9

### DIFF
--- a/roles/node_exporter/tasks/preflight.yml
+++ b/roles/node_exporter/tasks/preflight.yml
@@ -62,8 +62,6 @@
 
 - name: Gather currently installed node_exporter version (if any)
   ansible.builtin.command: "{{ _node_exporter_binary_install_dir }}/node_exporter --version"
-  args:
-    warn: false
   changed_when: false
   register: __node_exporter_current_version_output
   check_mode: false


SuperQ (approved):

SuperQ:
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

cudevmaxwell:
```
args:
    warn: false
```
https://docs.ansible.com/ansible/2.9/modules/command_module.html#parameter-warn

I believe this disabled the task returning a message like "Are you sure you want to use a command here, and not a module?"

```
ignore_errors: true
```

allows the playbook to continue on failure, which is necessary if node_exporter has not been installed yet. 


This PR allows the node_exporter role to work in Ansible 2.14+, but it does impact users using older Ansible versions.

Should `ignore_errors` always be present, no matter the Ansible version, and `warn: false` only be used for older Ansible versions?

cudevmaxwell:
```
args:
      warn: "{{ (ansible_version.full is version('2.14.0', '<')) | ternary('false', omit) }}"
```

I think something like this might work. If the Ansible version is less than 2.14, pass false to warn. If it's above 2.14, pass `omit`, which shouldn't set `warn` to anything, avoiding the error in newer Ansible versions. Still need to test if that will work.

cudevmaxwell:
Even better. `warn` wasn't really necessary in this case anyway. Tested on Ansible core 2.12.7, and the `command` module doesn't fire any warnings when used this way. 

Even back in 2.9: https://github.com/ansible/ansible/blob/5602bc2a8d492de2cb151d61a73ee4b409b1831f/lib/ansible/modules/commands/command.py#L183, it doesn't seem like running `node_exporter` this way would fire a warning. So that line can just be removed.

SuperQ:
Yup, this is likely an issue with much older versions. Many of these roles were developed long ago.

SuperQ (approved):

PR #14, opened 2022-12-23
node_exporter: Support filesystems mounted under '/home'

### LINKED ISSUE #13: node_exporter: Systemd's PreventHome causes error if filesystem is mounted under '/home'
In the node_exporter.service.j2 template, `ProtectHome` is set to `read-only` instead of `yes` only if `/home` is a separate partition. However, it is possible that a filesystem might be mounted under `/home` instead of at `/home`. Because `ProtectHome` is set to `yes` in that case, [node_exporter can't run statfs() on that filesystem](https://github.com/prometheus/node_exporter/blob/996563f9729588660d5d32826c46b0d7169b1334/collector/filesystem_linux.go#L80). 

(Copy of https://github.com/cloudalchemy/ansible-node-exporter/issues/271)

cudevmaxwell:
Set systemd's ProtectHome to 'read-only' instead of 'yes' if there exists filesystems which are mounted at or under '/home', instead of at '/home' exclusively.

Fixes #13.

Signed-off-by: Kevin Bowrin <kevinbowrin@cunet.carleton.ca>

### DIFF
--- a/roles/node_exporter/templates/node_exporter.service.j2
+++ b/roles/node_exporter/templates/node_exporter.service.j2
@@ -34,11 +34,11 @@ Restart=always
 RestartSec=1
 StartLimitInterval=0
 
-{% for m in ansible_mounts if m.mount == '/home' %}
-ProtectHome=read-only
-{% else %}
-ProtectHome=yes
+{% set protect_home = 'yes' %}
+{% for m in ansible_mounts if m.mount.startswith('/home') %}
+{%   set protect_home = 'read-only' %}
 {% endfor %}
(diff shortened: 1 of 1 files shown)


cudevmaxwell (commented):
I did it this way instead of the simpler fix suggested on the [original issue](https://github.com/cloudalchemy/ansible-node-exporter/issues/271) because it's possible that multiple filesystems might be mounted under /home, leading to multiple ProtectHome lines.

SuperQ (approved):
LGTM, thanks!

PR #17, opened 2023-01-13
feat: role/collection testing with ansible-test and molecule

gardar:
This PR enables testing of the collection and has a few fixes to make sure the collection passes all tests.

Collection sanity and integration tests are performed by ansible-test 
The roles are tested with molecule via ansible-test.
I decided to use ansible-test to run molecule to streamline the testing in the collection and to get the benefit of making ansible-test take care of testing against different ansible versions and python versions, also there is a official ansible-test github action when there is no official molecule action.

Most of the molecule config has been combined into one as duplicated definitions are no longer necessary after the roles got converted into a collection, with that being said, there is still room for even further merging of tests.

The molecule platforms have been switched over to the [dokken](https://github.com/test-kitchen/dokken-images) containers since those are actively maintained and I had some issues running the @paulfantom in the github ci.

The github action looks up what versions are supported in the role  (from [meta/runtime.yml](meta/runtime.yml)) and runs the tests against those versions, currently versions `2.9` - `2.14` are listed.

The github action also detects what role has been changed and then runs tests for all molecule scenarios for that role (and each test is then ran against all of the supported ansible versions)
For example if the `node_exporter` role is changed then it results in creating and executing the following tests:

```
node_exporter-latest-stable-2.9
node_exporter-latest-stable-2.10
node_exporter-latest-stable-2.11
node_exporter-latest-stable-2.12
node_exporter-latest-stable-2.13
node_exporter-latest-stable-2.14
node_exporter-alternative-stable-2.9
node_exporter-alternative-stable-2.10
node_exporter-alternative-stable-2.11
node_exporter-alternative-stable-2.12
node_exporter-alternative-stable-2.13
node_exporter-alternative-stable-2.14
node_exporter-default-stable-2.9
node_exporter-default-stable-2.10
node_exporter-default-stable-2.11
node_exporter-default-stable-2.12
node_exporter-default-stable-2.13
node_exporter-default-stable-2.14
```


SuperQ (approved):
LGTM.

SuperQ:
@paulfantom any final thoughts?

PR #20, opened 2023-01-19
Update install.yml

a-savinov:
Fix node_exporter download url 

Signed-off-by: Alexey Savinov <belsat.net@gmail.com>


SuperQ:
I think this was fixed in #17.

gardar:
> I think this was fixed in #17.

PR #25, opened 2023-02-12
Update selinux.yml

flortsch:
Fix failing check for redhat os family.

Signed-off-by: Florian Latifi <mail@florian-latifi.at>


gardar:
obsoleted by #57

PR #30, opened 2023-02-24
fix: writable token in all pull request workflows

gardar:
Accidentally only fixed one workflow trigger in #29 , this fixes the rest of the workflows.


SuperQ (approved):
LGTM

PR #37, opened 2023-03-02
fix: ansible 2.9 workaround for galaxy install from git

gardar:
Workaround for installing collection requirements from git on ansible 2.9 since ansible 2.9 does not support that.
Fixes regression introduced in #33


tests/integration/molecule.sh
```diff
@@ -23,10 +23,21 @@ fi
 # Install ansible version specific requirements
 if [ "$(printf '%s\n' "2.12" "$ansible_version" | sort -V | head -n1)" = "2.12" ]; then 
        python -m pip install molecule molecule-plugins[docker]
-else
+elif [ "$(printf '%s\n' "2.10" "$ansible_version" | sort -V | head -n1)" = "2.10" ]; then
```
SuperQ, line 26: Lol, bash hacks

SuperQ (approved):

PR #43, opened 2023-03-05
docs: Quote value latest to be explicit about the possible value for node_exporter_version

andygrunwald:
The value "latest" was written as a regular string in the docs. Quoting it via `marks it explicitly as a possible value to accept for node_exporter_version. Other wise, it is easily overlooked as an acceptable value.


SuperQ (approved):
Thanks

PR #50, opened 2023-03-06
fix: various role argument specs

gardar:
- Moves variables referenced from other variables out of tasks and into defaults.
- Adds missing argument specs for binary/checksum urls #28 
- Removes element validation from dict since it's only available on lists.


chriscarpenter12:
I can confirm this fixed the same reported issue I had running the node_exporter role. Tested on 0.1.5. Thanks @gardar

PR #52, opened 2023-03-06
refactor: avoid using command module for systemd version fact

gardar:
- Avoids using the command module and fragile regex parsing to get systemd version by getting systemd version from `package_facts`.
- This also removes the need for skipping the linting which caused the yaml linter to stop processing the rest of the preflight tasks, see bug: https://github.com/ansible/ansible-lint/issues/3139


gardar:
ping @SuperQ

SuperQ (approved):

PR #53, opened 2023-03-06
fix: replace eol platforms with current

gardar:
- Drops support for EOL platforms (distros)
- Adds support for current (latest currently supported) platforms.

### DIFF
--- a/.config/molecule/config.yml
+++ b/.config/molecule/config.yml
@@ -10,6 +10,12 @@ platforms:
     privileged: true
     cgroup_parent: docker.slice
     command: /lib/systemd/systemd
+  - name: almalinux-9
+    image: dokken/almalinux-9
+    pre_build_image: true
+    privileged: true
+    cgroup_parent: docker.slice
+    command: /lib/systemd/systemd
   - name: centos-7
     image: dokken/centos-7
     pre_build_image: true
@@ -18,6 +24,12 @@ platforms:
     command: /usr/lib/systemd/systemd
   - name: centos-stream-8
     image: dokken/centos-stream-8
+    pre_build_image: true
+    privileged: true
+    cgroup_parent: docker.slice
+    command: /lib/systemd/systemd
+  - name: centos-stream-9
+    image: dokken/centos-stream-9
     pre_build_image: true
     privileged: true
     cgroup_parent: docker.slice
--- a/.config/molecule/config.yml
+++ b/.config/molecule/config.yml
@@ -58,6 +58,12 @@ platforms:
     privileged: true
     cgroup_parent: docker.slice
     command: /lib/systemd/systemd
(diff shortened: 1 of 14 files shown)


SuperQ (commented):
Just as a small optimization, can we drop Ansible 2.11 as an EOL ansible version, just to cut down on the number of CI pipelines that run?

gardar:
> Just as a small optimization, can we drop Ansible 2.11 as an EOL ansible version, just to cut down on the number of CI pipelines that run?

Yes I believe so, these are the current ansible versions from the distro repos along with EOL dates.
None of them have ansible 2.11

| distro       | ansible package | version | EOL  | 
| ------------ | --------------- | ------- | ---- |
| almalinux-8  | ansible-core    | 2.13    | 2029-03-01 |
| almalinux-9  | ansible-core    | 2.13    | 2032-05-31 |
| centos-7     | ansible         | 2.9     | 2024-06-30 |
| centos-8-stream | ansible-core | 2.14    | 2024-05-31 |
| centos-9-stream | ansible-core | 2.14    | 2027-05-31 |
| debian-9     | ansible         | 2.7     | 2022-07-01 |
| debian-10    | ansible         | 2.9     | 2024-06-30 |
| debian-11    | ansible         | 2.10    | 2026-06-30 |
| fedora-30    | ansible         | 2.9     | 2020-05-26 |
| fedora-36    | ansible-core    | 2.12    | 2023-05-16 |
| fedora-37    | ansible-core    | 2.14    | 2023-12-15 |
| ubuntu-16.04 | ansible         | 2.0     | 2026-04-02 |
| ubuntu-18.04 | ansible         | 2.19    | 2028-04-01 |
| ubuntu-20.04 | ansible         | 2.9     | 2030-04-02 | 
| ubuntu-22.04 | ansible / ansible-core | 2.10 / 2.12    | 2032-04-09 |

SuperQ:
One thing about those EoL dates. I don't think it's worth supporting "extended support" timelines. For example we can drop 16.04, and soon 18.04 under the standard 5 year support cycle.

gardar:
> One thing about those EoL dates. I don't think it's worth supporting "extended support" timelines. For example we can drop 16.04, and soon 18.04 under the standard 5 year support cycle.

I'm torn between dropping the ELTS and keeping it. While you most likely wouldn't and should not really install prometheus/alertmanager on such old platforms, you might want to monitor your old systems with the exporters, and perhaps it's even more important to keep an eye on those old systems than the newer ones.

SuperQ:
It's more about what time we have to support such old platforms. We can say "May work, use at your own risk". We can avoid explicitly breaking old platforms. But officially supporting very old platforms is a drain on our limited resources.

SuperQ (approved):

PR #55, opened 2023-03-07
enhancement: Check that TLS cert and key can be read

volcan01010:
## Summary

This pull request adds an extra assertion to the preflight checks to confirm that the `node_exporter` process has read permission on the TLS key and certificate.

## Background

When we tried setting up the node exporter with TLS we found that the playbook ran successfully but that the node exporter was not running on the target server.
The error messages from `systemctl status node_exporter` did not contain useful information.
Eventually we traced the problem to the TLS certificate having `r-------- root root` permissions and therefore not being accessible to the `node-exp` system account.
Adding this extra check will prevent this happening to others.

## To test

I have not updated the `molecule` tests as there were no equivalent tests on the other preflight assertions. To trigger the assertion manually, you can run `chmod 0000 /path/to/your/tls.key` and then apply the role. It should fail during the preflight tests.

Thank you for your work on this role!

Note: This merge request has been migrated from the previous repository at https://github.com/cloudalchemy/ansible-node-exporter/pull/264

### DIFF
--- a/roles/node_exporter/tasks/preflight.yml
+++ b/roles/node_exporter/tasks/preflight.yml
@@ -41,7 +41,7 @@
         path: "{{ node_exporter_tls_server_config.cert_file }}"
       register: __node_exporter_cert_file
       become: true
-      become_user: "{{ _node_exporter_system_user }}"
+      become_user: "{{ node_exporter_system_user }}"
 
     - name: Check existence of TLS key file
       ansible.builtin.stat:
--- a/roles/node_exporter/tasks/preflight.yml
+++ b/roles/node_exporter/tasks/preflight.yml
@@ -62,7 +62,7 @@
           - "{{ __node_exporter_cert_file.stat.readable }}"
           - "{{ __node_exporter_key_file.stat.readable }}"
         msg: >-
-          {{ _node_exporter_system_user }} system user does not have read
+          {{ node_exporter_system_user }} system user does not have read
           permission on the TLS key and/or certificate.
 
 - name: Check if node_exporter is installed
--- a/roles/node_exporter/tasks/preflight.yml
+++ b/roles/node_exporter/tasks/preflight.yml
@@ -48,7 +48,7 @@
         path: "{{ node_exporter_tls_server_config.key_file }}"
       register: __node_exporter_key_file
       become: true
-      become_user: "{{ _node_exporter_system_user }}"
+      become_user: "{{ node_exporter_system_user }}"
 
     - name: Assert that TLS key and cert are present
       ansible.builtin.assert:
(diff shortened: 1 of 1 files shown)


gardar:
@volcan01010 Please fix CI errors

volcan01010:
Hi, the CI is passing now.

roles/node_exporter/tasks/preflight.yml
```diff
@@ -40,18 +40,31 @@
...
       register: __node_exporter_cert_file
+      become: true
+      become_user: "{{ _node_exporter_system_user }}"
 
     - name: Check existence of TLS key file
       ansible.builtin.stat:
         path: "{{ node_exporter_tls_server_config.key_file }}"
       register: __node_exporter_key_file
+      become: true
+      become_user: "{{ _node_exporter_system_user }}"
```
gardar, line 51:
```suggestion
      become_user: "{{ node_exporter_system_user }}"
```

roles/node_exporter/tasks/preflight.yml
```diff
@@ -40,18 +40,31 @@
...
           - "{{ __node_exporter_cert_file.stat.exists }}"
           - "{{ __node_exporter_key_file.stat.exists }}"
 
+    - name: Assert that node_exporter can read TLS key and cert
+      ansible.builtin.assert:
+        that:
+          - "{{ __node_exporter_cert_file.stat.readable }}"
+          - "{{ __node_exporter_key_file.stat.readable }}"
+        msg: >-
+          {{ _node_exporter_system_user }} system user does not have read
```
gardar, line 65:
```suggestion
          {{ node_exporter_system_user }} system user does not have read
```

gardar (changes requested):
#65

roles/node_exporter/tasks/preflight.yml
```diff
@@ -40,18 +40,31 @@
       ansible.builtin.stat:
         path: "{{ node_exporter_tls_server_config.cert_file }}"
       register: __node_exporter_cert_file
+      become: true
+      become_user: "{{ _node_exporter_system_user }}"
```
gardar, line 44:
```suggestion
      become_user: "{{ node_exporter_system_user }}"
```

gardar (changes requested):

gardar:
Looking at this I don't think this approach is totally correct, you are trying to become a user that does not exist yet. The node_exporter user is created in the install steps that are executed after the preflight.
(It would work fine if you already executed the role, or manually created the user, but fail in every other case.)

It could be a bit tricky to achieve this. You would need to check both the name of the owner/group of the file and the permissions of the file, and make sure the node exporter user falls within that scope.

volcan01010:
Ah, yes.  Good point.  I made the changes to the role on my own machine after previously running the script and creating the user.

This request was only meant to be a small tweak to provide a minor improvement, and I am not able to do much deeper work.  I will close the request.

Thanks for your time reviewing the request and to the wider team for keeping the Node Exporter role alive.

PR #62, opened 2023-03-09
feat: add systemd exporter role

gardar:


roles/systemd_exporter/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+systemd_exporter_version: 0.4.0
+systemd_exporter_binary_local_dir: ""
+systemd_exporter_binary_url: "https://github.com/povilasv/systemd_exporter/releases/download/v{{ systemd_exporter_version }}/\
```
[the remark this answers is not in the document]
gardar: Nice!

SuperQ (approved):

roles/systemd_exporter/meta/main.yml
```diff
@@ -0,0 +1,32 @@
...
+        - stretch
+        - buster
+    - name: EL
+      versions:
+        - '7'
+        - '8'
+    - name: Fedora
+      versions:
+        - '30'
+        - '31'
```
SuperQ, line 24: Non-blocking note: We should probably sync these lists with the platforms we're currently testing with.

gardar: Yep, planning to update this for all the roles in #53

PR #63, opened 2023-03-09
fix: policycoreutils python package name

mikaellanger:
I have a few legacy hosts with Centos 7 where trying to install python-policycoreutils fails.

At least in Centos 7 the packages is named policycoreutils-python, and finding no such package here either https://pkgs.org/download/python-policycoreutils

### DIFF
--- a/roles/alertmanager/tasks/selinux.yml
+++ b/roles/alertmanager/tasks/selinux.yml
@@ -1,7 +1,7 @@
 ---
 - name: Install selinux python packages [RedHat]
   ansible.builtin.package:
-    name: "{{ ['libselinux-python', 'python-policycoreutils']
+    name: "{{ ['libselinux-python', 'policycoreutils-python']
            if ansible_python_version is version('3', '<') else
            ['python3-libselinux', 'python3-policycoreutils'] }}"
     state: present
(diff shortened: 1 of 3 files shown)


SuperQ (changes requested):
Two things. This task is used in several roles, so should update all them.

I also think this is going to be a problem for redhat 8+, since this package has been renamed.

SuperQ:
Hmm, I guess if python3 is a good enough signal, it should work.

gardar:
The definition is present in these files:

https://github.com/prometheus-community/ansible/blob/d389ad29e203ed24aebb02796958837bdca6d35e/roles/prometheus/vars/main.yml#L12-L14
https://github.com/prometheus-community/ansible/blob/d389ad29e203ed24aebb02796958837bdca6d35e/roles/alertmanager/tasks/selinux.yml#L4-L6
https://github.com/prometheus-community/ansible/blob/d389ad29e203ed24aebb02796958837bdca6d35e/roles/node_exporter/tasks/selinux.yml#L4-L6

The package change came with python3 on redhat based platforms, that's why I thought the cleanest approach would be to make the conditional work on the python interpreter. (and yes you can blame me for the package name typo - whoops :) )

mikaellanger:
Changed the other two places.

> The package change came with python3 on redhat based platforms, that's why I thought the cleanest approach would be to make the conditional work on the python interpreter.

Based on available packages I agree that seems to be correct, yes.

gardar (approved):

gardar:
@mikaellanger please fix sign-off

SuperQ (approved):
Just need to fix the DCO.

mikaellanger:
Done

PR #65, opened 2023-03-09
refactor: underscore variable prefix should be reserved...

gardar:
...for internal variables that are not user configurable

While it's not specified in any standard that I'm aware of, it's common that variables that are used internally in a role are prefixed with a underscore.

These variables were once not user configurable, so it makes sense that they were prefixed in the past, but I don't see any reason to keep them that way.

### DIFF
--- a/roles/node_exporter/defaults/main.yml
+++ b/roles/node_exporter/defaults/main.yml
@@ -26,7 +26,6 @@ node_exporter_enabled_collectors:
 
 node_exporter_disabled_collectors: []
 
-# Internal variables.
-_node_exporter_binary_install_dir: "/usr/local/bin"
-_node_exporter_system_group: "node-exp"
-_node_exporter_system_user: "{{ _node_exporter_system_group }}"
+node_exporter_binary_install_dir: "/usr/local/bin"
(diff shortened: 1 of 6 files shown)


SuperQ:
We had an opinion that these vars were not supposed to be used by users. They used to be internal only. @paulfantom What do you think?

gardar:
> We had an opinion that these vars were not supposed to be used by users. They used to be internal only. @paulfantom What do you think?

Then they should be placed in `vars/` and not `defaults/`
Looking at cloudalchemy repos I see that initially they were, but then got moved to defaults to allow users to overwrite them.
With that being said, I don't see any downsides of having them overwritten by users myself.

paulfantom (approved):
The initial idea was to structure variables following [a pattern known from python](https://peps.python.org/pep-0008/#descriptive-naming-styles). Initially, those variables were placed in `vars/`. However, over time, people started using them and variables were moved into `defaults/` while keeping naming to preserve compatibility. Since we now moved the role to a different place and those variables are now meant to be overridden when needed I don't see any reason to keep prefixing them with underscores.

SuperQ (approved):

PR #66, opened 2023-03-10
skip_changelog: Don't run CI on push

SuperQ:
Avoid running ansible-ci workflow on push, since it doubles the number of executions for PRs built on in-repo branches.


gardar (approved):
LGTM

PR #67, opened 2023-03-12
Fixup version bumper job

SuperQ:
Move role repo discovery to a script to avoid shell escaping issues.


discover_role_repos.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+for defaults_file in roles/*/defaults/main.yml ; do
+  role="$(echo "${defaults_file}" | cut -f2 -d'/')"
+  yq eval "[{\"repo\": .${role}_repo, \"role\": \"${role}\"}]" "${defaults_file}"
+done | yq -o json -I=0
```
gardar, line 6: Shouldn't this be `vars/main.yml` now that the repo var lives there?

SuperQ: Yes, forgot to update that.

gardar (commented):
I think we should avoid polluting the root of the repo with ci related scripts and I suggest moving `discover_role_repos.sh -> .github/scripts/discover_role_repos.sh`
Which is the suggested script location from [github](https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#adding-scripts-to-your-workflow)

(And we should move `version_updater.sh` too)

SuperQ:
Done, moved scripts.

gardar (approved):

PR #69, opened 2023-03-13
docs: add workflow for auto generating ansible docs

gardar:
Using worklows from: https://github.com/ansible-community/github-docs-build

See: https://github.com/ansible-community/github-docs-build/wiki/Getting-started-with-GitHub-Pages

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -31,8 +31,8 @@ jobs:
       init-html-short-title: Prometheus.Prometheus Collection Docs
       init-extra-html-theme-options: |
           documentation_home_url=https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}/branch/main/
-        render-file-line: '> * `$<status>` [$<path_tail>](https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}/pr/${{ github.event.number }}/$<path_tail>)'
-        extra-collections: community.crypto amazon.aws community.dns community.sops
+      render-file-line: '> * `$<status>` [$<path_tail>](https://${{ github.repository_owner }}.github.io/\
+                         ${{ github.event.repository.name }}/pr/${{ github.event.number }}/$<path_tail>)'
 
     publish-docs-gh-pages:
       # for now we won't run this on forks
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
Nice

.github/workflows/docs-pr.yml
```diff
@@ -0,0 +1,93 @@
+---
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
```
bboreham, line 4: This license is not compatible with the Apache licence for this repo; can you find another way to do it?

.github/workflows/docs-push.yml
```diff
@@ -0,0 +1,51 @@
+---
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
```
bboreham, line 4: Ditto.

bboreham (commented):
Hi, please can you comment on the licence incompatibility introduced here.

.github/workflows/docs-pr.yml
gardar: Thanks for bringing this up. I get where you're coming from with the licensing concerns. Just to clear things up, the GPL-licensed workflow we're using is only for testing. It doesn't get included in the Ansible collection we actually publish. So, I thought this setup wouldn't affect our project's Apache license since the workflow is separate from the code we distribute. But I'm all for making sure we're on the right track and happy to chat more about it if needed!

bboreham: Quoting from GPL 3:
> The "Corresponding Source" for a work in object code form means all
the source code needed to generate, install, and (for an executable
work) run the object code and to modify the work, including scripts to
control those activities.

I don't think it's fruitful for me to chat about it; please engage with the CNCF's licencing experts if you want to.

gardar: Interesting! So judging by that quote from GPL 3 it looks like we can't have the ansible collection Apache licensed as the code in the collection can't run without Ansible and Ansible is GPL licensed?

matthiasr: We are not distributing Ansible though.

CNCF has lawyers who think about these thing. As a CNCF project we are bound to the rules they lay out, including the "no GPL in the repositories" one. These rules are there so that users of CNCF projects to not have to worry about these fine points 😃

gardar: #332

.github/workflows/docs-push.yml
gardar: #332

Issue #72, opened 2023-03-16
Role mysqld_exporter is leaking the exporter mysql password


laurent-indermuehle:
`mysqld_exporter_password` is not documented in the Readme and should have no_log set to true in argument-spec:
https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec

Currently, the task `prometheus.prometheus.mysqld_exporter : Copy the mysqld_exporter config file` is leaking the password.

SuperQ:
Good catch, thanks. Forgot to include the no_log here.

PR #73, opened 2023-03-16
fix: Don't log config deployments

SuperQ:
Avoid logging on some config files may contain secrets.

Fixes: https://github.com/prometheus-community/ansible/issues/72


roles/alertmanager/tasks/configure.yml
```diff
@@ -17,6 +17,7 @@
     group: alertmanager
     mode: 0644
     validate: "{{ _alertmanager_binary_install_dir }}/amtool check-config %s"
+  no_log: "{{ 'true' if __testing is defined else 'false' }}"
```
laurent-indermuehle, line 20: Maybe I miss something, but shouldn't this be the opposite?

`'false' if __testing is defined else 'true'`

laurent-indermuehle (commented):
Because the variable is used in a template, arguments_spec won't work?

Because to me, it would be better to have the no_log on the variable instead of the whole task.
You wouldn't need to check the `__testing` vars and in case of issue on production the task will output the error instead of the "no_log" warning.

gardar (changes requested):
Rather than using our own variables I would prefer to use variables that are already present, some of these:
- `no_log: "{{ molecule_no_log }}"`
- `no_log: "{{ (lookup('env', 'MOLECULE_DEBUG')) }}"`
- `no_log: "{{ (lookup('env', 'CI')) }}"` (comes from github ci)
- etc.

gardar:
> Because the variable is used in a template, arguments_spec won't work?
> 
> Because to me, it would be better to have the no_log on the variable instead of the whole task. You wouldn't need to check the `__testing` vars and in case of issue on production the task will output the error instead of the "no_log" warning.

Ansible doesn't support `no_log` for individual parameters of tasks, so you either hide the logging of the whole task or not at all.
Ansible modules however can define individual parameters as no log, but that's unfortunately no help here.

roles/alertmanager/tasks/configure.yml
SuperQ: Yes, that logic was backwards. Fixed.

SuperQ:
Verified with a local verbose molecule test:


```
TASK [prometheus.prometheus.alertmanager : Copy alertmanager config] ***********
changed: [debian-11] => {
    "censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result",
    "changed": true
}
```

roles/mysqld_exporter/tasks/configure.yml
```diff
@@ -23,6 +23,7 @@
     owner: root
     group: root
     mode: 0644
+  no_log: "{{ 'true' if __testing is defined else 'false' }}"
```
gardar, line 26:
```suggestion
  no_log: "{{ 'false' if lookup('env', 'CI') else 'true' }}"
```

gardar (changes requested):

roles/mysqld_exporter/tasks/configure.yml
SuperQ: Whups, missed one.

gardar (approved):

PR #74, opened 2023-03-16
enhancement: add `skip_install` variables to various roles

chrisbra:
This is similar to prometheus_skip_install variable and does what the name says: when set, it won't try to re-download alertmanager binaries

thanks for considering

### DIFF
--- a/roles/alertmanager/README.md
+++ b/roles/alertmanager/README.md
@@ -19,6 +19,7 @@ All variables which can be overridden are stored in [defaults/main.yml](defaults
 | Name           | Default Value | Description                        |
 | -------------- | ------------- | -----------------------------------|
 | `alertmanager_version` | 0.21.0 | Alertmanager package version. Also accepts `latest` as parameter. |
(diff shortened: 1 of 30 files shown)


gardar:
Wouldn't it make sense to have `skip_install` variables for all of the roles?

chrisbra:
yes, quite possible

chrisbra:
let me amend the pr with additional `skip_install` variables for the other roles.

chrisbra:
should be there now

gardar (changes requested):
Looks like we're not quite there yet, please take a look at the failures in the tests.

chrisbra:
yeah, looks like I messed up one of those conditions. Should be fixed now (and will re-push with my `Sign-Offed By` added

chrisbra:
do you want me to rebase on top of main?

gardar (approved):
LGTM!

SuperQ (approved):

PR #79, opened 2023-03-24
feat: allow to set `node_exporter_textfile_dir` to `false`

astappiev:
It would be nice to have an opportunity to set `node_exporter_textfile_dir` to `false` or `no` in case I want to disable the collector. E.g.

```yml
- name: Install & configure node_exporter
  ansible.builtin.include_role:
    name: setup_node_exporter
  vars:
    node_exporter_version: 'latest'
    node_exporter_textfile_dir: no
```

### DIFF
--- a/roles/node_exporter/tasks/configure.yml
+++ b/roles/node_exporter/tasks/configure.yml
@@ -39,7 +39,7 @@
     group: "{{ node_exporter_system_group }}"
     recurse: true
     mode: u+rwX,g+rwX,o=rX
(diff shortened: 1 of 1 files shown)


SuperQ:
Setting it to empty string does the same thing.

astappiev:
Yes, but no so convenient. I had to look at sources to understand how to disable it

gardar:
> Yes, but no so convenient. I had to look at sources to understand how to disable it

Documentation improvements are welcome

Issue #78, opened 2023-03-24
Checksum validation fails if `latest` version requested


astappiev:
The error says:
```
TASK [prometheus.prometheus.node_exporter : Discover latest version] ***************************************************************************************************************************************************************************************************************
ok: [host] => {"ansible_facts": {"node_exporter_version": "1.5.0"}, "attempts": 1, "changed": false}

TASK [prometheus.prometheus.node_exporter : Get checksum list from github] *********************************************************************************************************************************************************************************************************
fatal: [host]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'url'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Received HTTP error for https://github.com/prometheus/node_exporter/releases/download/vlatest/sha256sums.txt : HTTP Error 404: Not Found. Received HTTP error for https://github.com/prometheus/node_exporter/releases/download/vlatest/sha256sums.txt : HTTP Error 404: Not Found"}
```

As seen from log above, it retrieved version correctly, but somehow failed to override the variable.

astappiev:
The roots of this issue https://github.com/ansible/ansible/issues/22025
Perhaps, we can set another variable and check if it was set or not

Sylvain-AR:
Same issue here with ansible 2.14.3 on Debian and a simple playbook :
```
- name: Testing
  hosts: sys
  become: yes
  gather_facts: yes
  tasks:
    - ansible.builtin.ping:
  roles:
    - role: ...
    - role: ...
    - role: prometheus.prometheus.node_exporter
      node_exporter_version: "latest"
      node_exporter_web_listen_address: "0.0.0.0:9100"
      node_exporter_web_telemetry_path: "/metrics"

```

PR #80, opened 2023-03-24
feat: use separate var to store discovered version

astappiev:
Should close https://github.com/prometheus-community/ansible/issues/78, by introducing a new variable which stores discovered version, with a default fallback to the original variable.

### DIFF
--- a/roles/node_exporter/defaults/main.yml
+++ b/roles/node_exporter/defaults/main.yml
@@ -1,6 +1,5 @@
 ---
 node_exporter_version: 1.1.2
-_node_exporter_version: "{{ node_exporter_version }}"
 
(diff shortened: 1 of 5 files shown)


daniele-athome:
@astappiev you deleted the repository so this PR was closed automatically. Would you mind sharing the reasons for that? Did you find another fix/workaround?
Thanks!

astappiev:
Oh, I expected it was merged or so. Just cleaned the useless forks I have :/
Right now in my playbooks, I'm using [custom collection to download binaries from Github](https://github.com/astappiev/ansible-github). But it would be too complicated for simple use cases, feel free to use `Files changed` tab and copy my changes.

PR #83, opened 2023-04-05
enhancement: support ansible-vaulted basic auth passwords

iisojunn:
If ansible-vaulted password is passed to password_hash filter, it will throw and error: 'secret must be unicode or bytes, not ansible.parsing.yaml.objects.AnsibleVaultEncryptedUnicode' Handling the vaulted password with string filter first will avoid the problem.


gardar (approved):

iisojunn:
Rebased to add signed-off-by line (fix DCO check)

gardar:
The lint failed on a unrelated error which I just fixed in #84 
Can you please rebase?

PR #85, opened 2023-04-11
docs: avoid maintaining variable documentation in many places

gardar:
No need to maintain variable documentation in README when we got our online docs which are automatically generated.


SuperQ (approved):
Nice

PR #87, opened 2023-04-17
fix: tags support for included tasks

### LINKED ISSUE #86: Tags are not working
I wanted to quickly update my Prometheus config using the tag "prometheus_config". The playbook didn't work because the role uses "include_tasks".

I see 2 solutions:
1) Convert `include_tasks` to `import_tasks`.
2) Add `apply: {tags: ...}` to all ìnclude_tasks` call.

Is there a reason why this collection prefer `include_tasks`?

laurent-indermuehle:
This is a PR to fix #86

I think `import_tasks` should be prefered. But I don't want to impose my choice.

### DIFF
--- a/changelogs/fragments/fix86_apply_tags.yml
+++ b/changelogs/fragments/fix86_apply_tags.yml
@@ -1,3 +1,0 @@
----
-bugfixes:
-  - all roles - Fix tags to enable the execution of a specific tasks file as intended.
--- a/changelogs/fragments/fix86_apply_tags.yml
+++ b/changelogs/fragments/fix86_apply_tags.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - all roles - Fix tags to enable the execution of a specific tasks file as intended.
--- a/roles/alertmanager/tasks/main.yml
+++ b/roles/alertmanager/tasks/main.yml
@@ -1,6 +1,12 @@
 ---
 - name: Preflight
-  ansible.builtin.include_tasks: preflight.yml
+  ansible.builtin.include_tasks:
+    file: preflight.yml
+    apply:
+      tags:
+        - alertmanager_install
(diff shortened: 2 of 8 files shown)


gardar:
Thanks for the PR.
You are correct that currently the tags do very little, since they only apply to the include itself and the tags are missing from the included tasks.
The roles previously used the now deprecated `include` module and I had not tested the tags after migrating to `include_tasks`.

Rather than applying the tags to everything I would prefer that the tags were added to the relevant tasks instead, as applying the tags to the whole import/include will run some tasks that don't necessarily need to run.

changelogs/fragments/fix86_apply_tags.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - all roles - Fix tags to enable the execution of a specific tasks file as intended.
```
gardar, line 3: This is not needed as the changelog is automatically generated from the PR title, please update the title with the relevant tag.
See here: https://github.com/prometheus-community/ansible/blob/main/CONTRIBUTING.md#releases

gardar (changes requested):

laurent-indermuehle:
I discovered, thanks to the bot, that my commits must be signed-off. I just did. 

I think the contributors guide is lacking this information. But I appreciate the effort you make to automate quality control.

Ready for review.

laurent-indermuehle:
I fixed a yaml indentation (thanks to ansible-lint, I love this tool). But I got trapped again by the signed-off of my commit. This is not documented in the contributors guide either.
Could you please remove this constrain or explain how to configure this repository to always signed-off commits?

Also, I did what your requested in your code review and resolved the comment, but this PR is still marked "requested changes". Do you know why?

gardar (approved):

PR #88, opened 2023-04-19
fix: mysqld_exporter should actually respect the mysqld_exporter_host variable

janost:
This change fixes 2 issues.

The first issue is setting the `mysqld_exporter_host` variable to anything will be ignored and `localhost` will be used instead.

The second issue is even if `mysqld_exporter_host` is set, the `socket=...` line will be appended (since `mysqld_exporter_socket` has a default value) , forcing mysqld_exporter to use that, essentially ignoring the host and port settings.


gardar:
Can you please rebase? The CI failures were fixed in #53

SuperQ (approved):

PR #90, opened 2023-04-25
fix: lookup('url', ..) causes known issue on MacOS controllers

rjlasko:
This PR addresses a [known issue](https://github.com/ansible/ansible/issues/76322) running `lookup('url', ..)` on MacOS.

The writeup in the issue indicates that there are 2 possible solutions:
1. uninstall requests and urllib3
2. use the uri module instead of the lookup

This ticket pursues the latter option, as the former is unrealistic for many users.

Additionally, this work includes downloading the node_exporter archive into a temporary directory, which is cleaned up after copying the extracted binary to the target.

### DIFF
--- a/roles/node_exporter/tasks/install.yml
+++ b/roles/node_exporter/tasks/install.yml
@@ -17,7 +17,7 @@
     home: /
   when: node_exporter_system_user != "root"
 
-- name: Discover latest version
+- name: Install from non-local archive
   when:
     - node_exporter_binary_local_dir | length == 0
     - not node_exporter_skip_install
--- a/roles/node_exporter/tasks/install.yml
+++ b/roles/node_exporter/tasks/install.yml
@@ -22,18 +22,12 @@
     - node_exporter_binary_local_dir | length == 0
     - not node_exporter_skip_install
   block:
-    - name: create temporary directory for download
-      become: false
-      ansible.builtin.tempfile:
-        state: directory
-      register: __node_exporter_tempdir
-      delegate_to: localhost
 
     - name: Download node_exporter binary to local folder
       become: false
       ansible.builtin.get_url:
         url: "{{ node_exporter_binary_url }}"
-        dest: "{{ __node_exporter_tempdir.path }}/node_exporter-{{ node_exporter_version }}.linux-{{ go_arch }}.tar.gz"
+        dest: "/tmp/node_exporter-{{ node_exporter_version }}.linux-{{ go_arch }}.tar.gz"
         checksum: "sha256:{{ __node_exporter_checksum }}"
         mode: '0644'
       register: _download_binary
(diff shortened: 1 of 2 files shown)


gardar (commented):
Thanks, I like the tempfile solution, rather than using the hardcoded /tmp path.
However this introduces the unwanted side-effect that the binary will be downloaded to a local temp dir on each run of the role, rather than re-using the binary that's downloaded on the first run. 

I'm not sure about how I feel about using uri instead of the url lookup for the following reasons:
 - Wouldn't it be better to properly fix the controller? If you fix it you will have fixed the issue in any roles you may encounter, rather than adding workarounds to all of them.
 - Using the lookup looks cleaner and we can skip extra tasks.
 - The proposed change executes the version and checksum lookups on the target host rather than the controller, this will cause issues if your target hosts don't have internet access. This could be fixed by delegating the tasks to localhost, but won't that re-introduce the issue on mac os controllers? Or is the issue only present with the url lookup module and not the uri module on mac os?

Edit: Does `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` not fix the issue on mac os? It's mentioned in the link above and seems like the most sane solution of the 3.

rjlasko:
Hello @gardar.  Thanks for the quick feedback!

> the tempfile solution, rather than using the hardcoded /tmp path

I see issue that you are getting at.  My primary interest was to ensure that there was no trace left upon completion of the role.  Perhaps there is still a way to do this (handler?), but for now i will revert that commit as admittedly it was out of scope for the problem i was trying to address.

gardar:
> I see issue that you are getting at. My interest here was to ensure that there was no trace of the downloaded file left upon completion of the role. Perhaps there is still a way to do this (handler?), but for now i will revert that commit as admittedly it was out of scope for the problem i was trying to address.

Agreed, I'm not a fan of leaving temp files around after the run but there are still cases where someone might want to keep the downloaded binary to avoid downloading it in the next run. The download destination could perhaps be moved to a variable, so that you could choose where it would reside and then also have a optional cleanup task. 
But for now you can always just do the cleanup in your playbook.

> facepalm Well that is embarrassing. Of course you are right. These should (+will) be delegated to the controller for exactly the reasons you've suggested, and because running lookup on the a Mac controller was my issue in the first place.

Does the uri module work when delegated to localhost on mac? One would think that both the url lookup and the uri module delegated to localhost would behave in a similar way.

> One of my most closely held software design principles is that it should be sufficiently robust such that it helps its users not have to care about the nuances or deficiencies of their system, especially when those are dependencies of the software itself. This certainly doesn't mean that the software should try to cover every possible circumstance. Rather, it should be willing to take on reasonable burdens if they absolve the user, how ever many of them there may be and at whatever level of skill, from having to search for and create workarounds to use that software.

I do agree with that statement.
While the proposed change is fairly minimal I am still not fully convinced that it is within the scope of the collection to be making workarounds for different controller hosts, installing dependencies on the controller hosts or configuring them in any way.

It's a known fact that ansible is not fully cross platform and different controller hosts require different setup, the workaround for this issue and also other controller host issues are listed in the official ansible documentation: https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#running-on-macos
It looks to me like this should be a part of the setup for everyone that's running ansible on mac os.

If we were to add controller host specific code to the collection, then we would also need to be able to test it in the ci.

With that being said, the readme for the roles already lists dependencies that are required on the controller hosts.
https://github.com/prometheus-community/ansible/blob/f6e6ac041e22ec7c9cefaa14f8782e7f0236ddf7/roles/node_exporter/README.md?plain=1#L12-L13
Perhaps it would be a good idea to add the information about `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` there?

(I hope that I don't sound overly negative :cold_sweat: )

rjlasko:
> Does the uri module work when delegated to localhost on mac? One would think that both the url lookup and the uri module delegated to localhost would behave in a similar way.

I've confirmed that the `uri` module does not suffer the same issue as `lookup('url', ..)`.

PR #94, opened 2023-05-03
[bugfix] node_exporter: Fix Systemd ProtectHome option in service unit

cudevmaxwell:
Fixes an issue with the jinja2 snippet which is used to create the node_exporter Systemd unit. More details here: https://github.com/prometheus-community/ansible/issues/13

Jinja2 namespaces are used to ensure the variable `protect_home` can be set in the parent scope of the `for` loop looking through the mounts.


cudevmaxwell:
Some testing at the REPL:

```
>>> import jinja2
>>> t = """
... {% set ns = namespace(protect_home = 'yes') %}
... {% for m in ansible_mounts if m.mount.startswith('/home') %}
... {%   set ns.protect_home = 'read-only' %}
... {% endfor %}
... ProtectHome={{ ns.protect_home }}
... NoNewPrivileges=yes
... """
>>> tmpl = jinja2.Template(t)
>>> tmpl.render(ansible_mounts=[{"mount":"/root"}, {"mount":"/"}])
'\n\n\nProtectHome=yes\nNoNewPrivileges=yes'
>>> tmpl.render(ansible_mounts=[{"mount":"/root"}, {"mount":"/"}, {"mount":"/home/here"}])
'\n\n\n\n\nProtectHome=read-only\nNoNewPrivileges=yes'
>>> tmpl.render(ansible_mounts=[{"mount":"/home"}, {"mount":"/root"}, {"mount":"/"}])
'\n\n\n\n\nProtectHome=read-only\nNoNewPrivileges=yes'
```

SuperQ (approved):
Thanks, it'd be good to add a test for the correct ProtectHome value in the molecule tests.

cudevmaxwell:
I think I understand how to add a molecule test to ensure the service file exists or has the right contents (something like https://github.com/prometheus-community/ansible/blob/5a3968dd207a249b0841bf3e45000ac52f4ec991/roles/prometheus/molecule/alternative/tests/test_alternative.py#LL24C1-L35C21) but I'm not sure how to actually test this. Three machines would need to be spun up: 
- one with no special partitions, test that ProtectHome is set to 'yes'.
- one with a partition mounted at /home, test that ProtectHome is set to 'read-only'.
- one with a partition mounted at /home/some/other/path, test that ProtectHome is set to 'read-only'.

SuperQ:
Yea, I was looking for some simple pytest options to assert "line in file", but I didn't find anything reasonable in the docs.

Maybe @gardar knows this stuff better.

gardar:
> Yea, I was looking for some simple pytest options to assert "line in file", but I didn't find anything reasonable in the docs.
> 
> Maybe @gardar knows this stuff better.

Something like [`file.contains`](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.file.File.contains) ?

But perhaps [`systemd_properties`](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.service.Service.systemd_properties) would be even better.

gardar (approved):

Issue #92, opened 2023-04-26
node_exporter "Create textfile collector dir" task - Consistent "change" state


fitz123:
**Description**
The "Create textfile collector dir" task in the node_exporter role for the Prometheus Ansible collection is consistently reporting a "changed" state upon every run. [The task in question](https://github.com/prometheus-community/ansible/blob/main/roles/node_exporter/tasks/configure.yml#L34-L42)

Other system services generate their own metrics and place the corresponding metric files in the "{{ node_exporter_textfile_dir }}" directory. These services belong to the "{{ node_exporter_system_group }}" group, enabling them to change the group ownership of created metric files to the desired "{{ node_exporter_system_group }}". The node-exporter can read from these metric files, but the services are unable to change the ownership to the "{{ node_exporter_system_user }}". This results in a "changed" state for the "Create textfile collector dir" task after every run, as long as there are new or updated metric files within the "{{ node_exporter_textfile_dir }}" directory.

**Problem**
This issue is problematic because I monitor Ansible changes for security reasons and to detect any unexpected changes. Constant changes are undesirable and should not occur under normal circumstances.

Expected Outcome:
When all files and directories within the node_exporter_textfile_dir directory have the correct group set, no Ansible changes should be reported. The role should neither alter the "user" for files nor report any changes regarding this.

**Solution Proposal**
I have tried to come up with a suitable solution but have been unsuccessful. Unfortunately, there is no associated PR. Possible solutions could include:

1. Setting `recurse: false` - This would set the desired ownership for the directory itself, allowing the user to manage and verify that text metric files have the correct group, which node-exporter can read from.
2. Using `changed: false` - This option is also not ideal.

I am unsure how to address this issue and welcome any suggestions or proposals for a resolution.

mikelsid:
imo the solution with setting `recurse: false` should work, since, according to the [source](https://github.com/prometheus/node_exporter/blob/master/collector/textfile.go), collector only scrapes files in this directory, excluding subdirectories and g+rwX,o=rX guarantees, that all files created in the directory will have the desired group

gardar:
I suspect that the recursive option was set as a safety precaution to ensure that node_exporter can read the files in the directory.
But as you are experiencing there are different scenarios where node_exporter can very well read the directory contents, without enforcing the user/group/mode that the task sets.

A better solution would be to confirm that the user can read the contents, with the `ansible.builtin.stat` module and `become` and then only change the permissions if the node exporter user can't access the files.

[...]

PR #105, opened 2023-05-15
issue 92: set recurse -> false for textfile collector dir

fitz123:
because it causes consistent ansible changes in case there are new metric files created by other users
[ issue 92](https://github.com/prometheus-community/ansible/issues/92)


roles/node_exporter/tasks/configure.yml
```diff
@@ -37,7 +37,7 @@
     state: directory
     owner: "{{ node_exporter_system_user }}"
     group: "{{ node_exporter_system_group }}"
-    recurse: true
+    recurse: false  # https://github.com/prometheus-community/ansible/issues/92
```
gardar, line 40:
```suggestion
```

gardar (changes requested):
The default option for recurse is false, so we can just remove the parameter: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-recurse

Also, there is no need to link the PR as a comment since git does track it and you can just look at the commit history to see why it got removed.

gardar:
Linting errors were unrelated and have been fixed in main, please rebase.

gardar:
Err, a new version of ansible-lint was released this morning which brings us new linting issues. 
I'll try to fix it later today, appreciate the patience.

gardar:
Mind rebasing again? Linting issues fixed in #109

gardar (approved):

PR #106, opened 2023-05-15
minor: Add ansible 2.15 support

gardar:
- Adds support for Ansible 2.15
- CI: added support for triggering role tests manually.
  By default the roles are only tested if we detect any changes to the role files, but now if you add a label to the PR with the name of the role then that will trigger a test.


gardar:
@SuperQ @paulfantom
FYI: added support for triggering role tests manually.
By default the roles are only tested if we detect any changes to the role files, but now if you add a label to the PR in the form of `test-role-ROLENAME` then that will trigger a test of the role `ROLENAME`

PR #110, opened 2023-05-18
skip_changelog: More fixes to version bumper

SuperQ:
* Check for remote branch before trying to build a changeset.
* Fail exit if push fails.
* Add pull-request permission to workflow.

### DIFF
--- a/.github/workflows/version_bumper.yml
+++ b/.github/workflows/version_bumper.yml
@@ -27,6 +27,7 @@ jobs:
     needs: discover-role-repos
     permissions:
       contents: write
+      pull-request: write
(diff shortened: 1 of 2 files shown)


gardar (approved):

.github/workflows/version_bumper.yml
```diff
@@ -27,6 +27,7 @@ jobs:
     needs: discover-role-repos
     permissions:
       contents: write
+      pull-request: write
```
gardar, line 30: Whoops, approved and merged already before I realized it's `pull-requests` and not `pull-request`

PR #89, opened 2023-04-25
fix: Install package fact dependencies needs to be run as root

JonathanMbt:

PR #114, opened 2023-05-21
fix: `become` when installing packages in preflight

lpulley:


lpulley:
This is a dupe of #89, my bad!

PR #117, opened 2023-05-26
refactor: use range for supported ansible versions

gardar:
* Use version range for collection supported Ansible versions. Previous method for listing supported versions resulted in warnings on patch level versions.
* Adds CI script for properly parsing PEP440 version ranges.
* fixes: #144


tisba:
This won't work for ansible >2.15.0 (like the current 2.15.2), correct?

ducnm0711:
The collection does not work with patch level version:
```
$ ansible --version
ansible [core 2.13.11]
  config file = /etc/ansible/ansible.cfg
  python version = 3.8.10 (default, May 26 2023, 14:05:08) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True

$ ansible-playbook commons.yml -t node_exporter
[WARNING]: Collection prometheus.prometheus does not support Ansible version 2.13.11
 ```

theunkn0wn1:
Any plans on merging this soon?

SuperQ (approved):
Nice. I suppose later we could add an option to skip some specific versions in testing that we don't want to care about.

PR #127, opened 2023-05-31
feat: Add pushgateway role

SuperQ:
Add role to manage the Prometheus Pushgateway.


gardar (changes requested):
Don't forget to add the integration tests, to ensure the CI runs molecule.


EDIT: we can take care of that in the CI, see: #135

gardar (approved):

PR #137, opened 2023-05-31
fix: Skip config validation when not installing

tvenieris:
Otherwise validation fails because amtool binary is missing.


SuperQ:
Yea, this doesn't make sense to me as much. IIRC the skip install is meant for out-of-band installs and runs after installing. Not for blindly pushing a broken config.

PR #149, opened 2023-06-19
fix: Always Install amtool for alertmanager

tvenieris:
amtool is required for the validation of the configuration file so its installation is needed even when the variable alertmanager_skip_install is set to 'true'.

Signed-off-by: Thomas Venieris <thomas.venieris@gmail.com>


tvenieris:
I made this PR in response to: https://github.com/prometheus-community/ansible/pull/137

PR #150, opened 2023-06-19
feature: skip alertmanager service installation

tvenieris:


tvenieris:
The restart handler is not affected by this new variable, because an "alertmanager" service is created by another role that handles the container lifecycle, and we still want the restart to happen when there is a configuration change.

gardar:
Obsoleted by #425 

Please use `--skip-tags install` or `--skip-tags <role name>_install` to skip install

PR #155, opened 2023-07-11
fix: alertmanager deprecated PIDFile= path warning

atishaninov-xometry:
Starting from systemd 239 ([commit link](https://github.com/systemd/systemd/commit/a2d1fb882c4308bc10362d971f333c5031d60069)), it started showing warning about using deprecated path to pidfile in `/var/run` instead of `/run`.

This warning can be seen by launching command `systemctl status alertmanager` or sometring like `journalctl -xeu alertmanager`:

```/etc/systemd/system/alertmanager.service:12: PIDFile= references a path below legacy directory /var/run/, updating /var/run/alertmanager.pid → /run/alertmanager.pid; please update the unit file accordingly.```

### DIFF
--- a/roles/alertmanager/templates/alertmanager.service.j2
+++ b/roles/alertmanager/templates/alertmanager.service.j2
@@ -17,7 +17,11 @@ StartLimitIntervalSec=0
 
 [Service]
 Type=simple
+{% if (ansible_facts.packages.systemd | first).version is version('239', '>=') %}
+PIDFile=/run/alertmanager.pid
+{% else %}
 PIDFile=/var/run/alertmanager.pid
+{% endif %}
 User=alertmanager
 Group=alertmanager
 ExecReload=/bin/kill -HUP $MAINPID


gardar (approved):

gardar:
Thanks!
Do you know if using `/run/alertmanager.pid` will cause issues on versions older than 239? Wondering if we could just use `/run/` for all versions.

atishaninov-xometry:
As far as i know, `/run` directory exists on systemd-based systems from 2011: https://lwn.net/Articles/436012/
Since introduction of `/run`, many distros implemented bind-mounting `/var/run` to `/run` or vice versa for compatibility.
I think it's safe to use `/run` for all supported distros.

gardar:
> As far as i know, `/run` directory exists on systemd-based systems from 2011: https://lwn.net/Articles/436012/ Since introduction of `/run`, many distros implemented bind-mounting `/var/run` to `/run` or vice versa for compatibility. I think it's safe to use `/run` for all supported distros.

Sounds good, do you mind changing it to /run/ for all versions then?
Thanks.

PR #156, opened 2023-07-12
feat: add option to disable all default node_exporter collectors

dstevensson:
Hi,
I'm new to the project, kinda new to ansible and not very used to github. Let me know if something looks wrong.
This PR will add the ability disable all default collectors, allowing you to set an explicit list of enabled collectors when used together with `node_exporter_enabled_collectors`. The behavior is documented over at [node_exporter repo](https://github.com/prometheus/node_exporter#collectors).

I didn't update any tests as updating `alternative` would impact the behavior of  the var `node_exporter_disabled_collectors`.

### DIFF
--- a/roles/node_exporter/defaults/main.yml
+++ b/roles/node_exporter/defaults/main.yml
@@ -27,6 +27,8 @@ node_exporter_enabled_collectors:
 
 node_exporter_disabled_collectors: []
 
+node_exporter_disable_default_collectors: false
+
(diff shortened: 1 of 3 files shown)


SuperQ (commented):
This is unnecessary as you can disable the defaults like this:

```yaml
node_exporter_enabled_collectors:
  - disable-defaults
  - cpu
  - etc
```

dstevensson:
Ah, I didn't realize it was that easy. Thanks!

PR #163, opened 2023-07-20
fix(node_exporter): correct indent in systemd template

tigattack:
Before:
image

After:
image

### DIFF
--- a/roles/node_exporter/templates/node_exporter.service.j2
+++ b/roles/node_exporter/templates/node_exporter.service.j2
@@ -23,13 +23,13 @@ ExecStart={{ node_exporter_binary_install_dir }}/node_exporter \
 {% for collector in node_exporter_disabled_collectors %}
     '--no-collector.{{ collector }}' \
 {% endfor %}
-{% if node_exporter_tls_server_config | length > 0 or node_exporter_http_server_config | length > 0 or node_exporter_basic_auth_users | length > 0 %}
+{% if node_exporter_tls_server_config | length > 0 or node_exporter_http_server_config | length > 0 or node_exporter_basic_auth_users | length > 0 -%}
     {% if node_exporter_version is version('1.5.0', '>=') %}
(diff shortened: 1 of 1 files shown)


SuperQ:
Looks like this breaks the startup of the `alternative` test. The exporter fails to run.

roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -23,13 +23,13 @@ ExecStart={{ node_exporter_binary_install_dir }}/node_exporter \
 {% for collector in node_exporter_disabled_collectors %}
     '--no-collector.{{ collector }}' \
 {% endfor %}
-{% if node_exporter_tls_server_config | length > 0 or node_exporter_http_server_config | length > 0 or node_exporter_basic_auth_users | length > 0 %}
+{% if node_exporter_tls_server_config | length > 0 or node_exporter_http_server_config | length > 0 or node_exporter_basic_auth_users | length > 0 -%}
```
SuperQ, line 26: I think this part of the change is breaking the alternative test.

```suggestion
{% if node_exporter_tls_server_config | length > 0 or node_exporter_http_server_config | length > 0 or node_exporter_basic_auth_users | length > 0 %}
```

roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -23,13 +23,13 @@ ExecStart={{ node_exporter_binary_install_dir }}/node_exporter \
...
     {% if node_exporter_version is version('1.5.0', '>=') %}
     '--web.config.file=/etc/node_exporter/config.yaml' \
     {% else %}
     '--web.config=/etc/node_exporter/config.yaml' \
     {% endif %}
-{% endif %}
+{% endif -%}
```
SuperQ, line 32:
```suggestion
{% endif %}
```

gardar (changes requested):
Please rebase

PR #169, opened 2023-07-27
Fix mysqld_exporter world-readable secrets

tvenieris:
The mysqld_exporter config includes a MySQL user passsword and the web config may include htpasswd hashes. Don't allow everyone to peek them.


SuperQ (approved):
Thanks

PR #171, opened 2023-07-27
fix ansible-lint: risky-octal & no-same-owner

tvenieris:


SuperQ (approved):
Strange that the CI pipeline is not warning about these already. The `risky-octal` rule should be enabled by the `production` profile in our config.

PR #172, opened 2023-07-28
feature: blackbox exporter user/group configurable

tvenieris:
Also, binary and configuration owner is changed to root. User blackbox-exp should be prevented from modifying or deleting either the configuration or itself, according to the principle of least privilege.


SuperQ (approved):
Thanks

SuperQ:
Strange that the label trigger didn't happen on the first pass.

SuperQ:
Oops, forgot to update the `meta/argument_specs.yml`.

PR #173, opened 2023-07-28
feature: Make config installation dir configurable

tvenieris:


roles/blackbox_exporter/defaults/main.yml
```diff
@@ -65,3 +65,6 @@ blackbox_exporter_configuration_modules:
 #      preferred_ip_protocol: ip6
 #      validate_answer_rrs:
 #        fail_if_matches_regexp: [test]
+
+# Where to put the blackbox_exporter.yml main configuration file
+blackbox_exporter_conf_dir: /etc
```
SuperQ, line 70: Let's be consistent with `conf` vs `config` on all roles.

```suggestion
blackbox_exporter_config_dir: /etc
```

SuperQ (changes requested):

SuperQ:
This also needs to update the `meta/argument_specs.yml` in order to update the auto-generated role docs.

roles/blackbox_exporter/defaults/main.yml
tvenieris: Changes applied. Apologies for not knowing the purpose of the meta/argument_specs.yml file.

roles/blackbox_exporter/defaults/main.yml
```diff
@@ -68,3 +68,6 @@ blackbox_exporter_configuration_modules:
 #      preferred_ip_protocol: ip6
 #      validate_answer_rrs:
 #        fail_if_matches_regexp: [test]
+
+# Where to put the blackbox_exporter.yml main configuration file
+blackbox_exporter_config_dir: /etc
```
SuperQ, line 73: Bleh, we should probably fix this to match the other roles.

SuperQ (approved):

roles/blackbox_exporter/defaults/main.yml
SuperQ: Followup issue: https://github.com/prometheus-community/ansible/issues/178

PR #177, opened 2023-07-29
Add test for argument_specs matching

SuperQ:
Compare contents of `meta/arguments_spec.yml` against `defaults/main.yml` on each role to make sure keys match.


gardar:
Nice!

Perhaps we should take it a step further and generate the defaults from the argument_specs automatically?
Like what's being discussed here: https://github.com/ansible/ansible/issues/77664

gardar:
We just need to move the example content into the descriptions of the argument specs, we could then just take the argument spec descriptions and convert those into comments in the defaults.
But perhaps it's time to move the docs/examples out of the defaults?

PR #182, opened 2023-08-03
fix: Install content on the target machine instead of the provisioning

lykholat:
When trying to install node_exporter on the target machines, at a certain point the binaries would start to be installed on the machine where ansible is running which is not the point when doing automation tasks. This correction will avoid the installation of the exporter on the provisioning machine and also solve this error:
TASK [prometheus.prometheus.node_exporter : Download node_exporter binary to local folder] ************************
fatal: [...*]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ubuntu@localhost: Permission denied (publickey).", "unreachable": true}

Now, the binaries will be downloaded to ubuntu@targetmachine instead of ubuntu@localhost

### DIFF
--- a/roles/blackbox_exporter/tasks/install.yml
+++ b/roles/blackbox_exporter/tasks/install.yml
@@ -27,14 +27,13 @@
   until: _download_binary is succeeded
   retries: 5
   delay: 2
-  delegate_to: localhost
   check_mode: false
   when: not blackbox_exporter_skip_install
 
 - name: Propagate blackbox exporter binary
   ansible.builtin.copy:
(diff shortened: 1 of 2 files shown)


gardar:
There are reasons why it is downloaded on the provisioner and not directly on the target, it saves you from downloading the binary multiple times and it's not always that the target hosts can access the internet (although that is less of an issue after #28 got merged).

I understand why it might be appealing to be able to download the binaries directly on the target machines. But rather than changing the roles from fetching the binaries on the provisioner to fetch them on the target it would be better if it were a configurable option, so that you could choose which method you want.

With that being said, I think there might be something wrong with the setup on your provisioner, AFAIK ansible should use a local connection by default when talking to localhost and not use ssh.

gardar:
#425

Issue #183, opened 2023-08-04
consider not modifying the node exporter user if it already exists


bluppfisk:
Using the node_exporter role without fully checking all the steps (mea culpa), I was astonished to find out that it had reduced the user I wanted to run node-exporter as to a no-login, no-home, system user. This then caused other issues with other services running under this user's name and that very much expect to have a home directory of their own.

I think the role should check whether the user already exists and not modify the user if so.

The service file should then also not include `ProtectHome=yes` as this would effectively disable it from accessing its own home, where, in my case all the metrics prom files are written to.

SuperQ:
No, we want our Ansible roles to be idempotent. We need to maintain all configurations in order for this policy to be maintained.

`ProtectHome` is a different issue, we can add a check like we do for mounts to not enable `ProtectHome` if the textfile path contains `/home`.

bluppfisk:
I'll still need to use a fork for my own purposes, but I understand that this may not work for all. The latter part is a sweet and rapid improvement :) cheers

PR #184, opened 2023-08-04
fix(node_exporter): Fix ProtectHome for textfiles

SuperQ:
Set the node_exporter `ProtectHome=read-only` when the textfile dir is in `/home`.

Fixes: https://github.com/prometheus-community/ansible/issues/183

### DIFF
--- a/roles/node_exporter/templates/node_exporter.service.j2
+++ b/roles/node_exporter/templates/node_exporter.service.j2
@@ -42,6 +42,9 @@ StartLimitInterval=0
 {% for m in ansible_mounts if m.mount.startswith('/home') %}
 {%   set ns.protect_home = 'read-only' %}
 {% endfor %}
+{% if node_exporter_textfile_dir.startswith('/home') %}
+{%   set ns.protect_home = 'read-only' %}
+{% endif %}
 ProtectHome={{ ns.protect_home }}
 NoNewPrivileges=yes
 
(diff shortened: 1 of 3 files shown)


roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -42,6 +42,9 @@ StartLimitInterval=0
 {% for m in ansible_mounts if m.mount.startswith('/home') %}
 {%   set ns.protect_home = 'read-only' %}
 {% endfor %}
```
gardar, line 44: Doesn't this serve the same purpose as the proposed change?

gardar: Ok can we then remove this in favor of the proposed change or is there a scenario where the mountpoint `/home` needs to have protecthome without the textfile being there?

gardar: Aha gotcha!

gardar (approved):

SuperQ:
I have no idea what's causing the node_exporter alternative to fail. :(

gardar:
> I have no idea what's causing the node_exporter alternative to fail. :(

Found out that it was because the version number variable set for that scenario was set to version 1.0.0 which does not support `--web.config.file`
The way the version number variable was set caused it to be set only for the preparation task, which made the rest of the run unaware of the version being used.

Fixed in #197 by moving the molecule variables of all of the roles out of the playbooks and into the molecule configs so that they are accessible for every step of the molecule runs.
Also added logging on service status failures so that we can spot issues like this in the future without jumping through multiple hoops.

Issue #187, opened 2023-08-15
How to run Prometheus in agent mode?


hongbo-miao:
I am wondering how to pass `--enable-feature=agent` to allow Prometheus running in agent mode: https://prometheus.io/blog/2021/11/16/agent/

I didn't find the info https://github.com/prometheus-community/ansible/tree/main/roles/node_exporter#requirements including the links inside.

Any guide would be appreciate, thanks! 😃

PR #198, opened 2023-08-19
feature: Agent mode support

rooty0:
This PR adds support to run prometheus in [agent mode](https://prometheus.io/blog/2021/11/16/agent/).

Fixes: #187


roles/prometheus/templates/prometheus.service.j2
```diff
@@ -13,12 +13,16 @@ User=prometheus
 Group=prometheus
 ExecReload=/bin/kill -HUP $MAINPID
 ExecStart={{ _prometheus_binary_install_dir }}/prometheus \
+{% if not prometheus_agent_mode %}
```
rooty0, line 16: we can't use this arguments with agent mode, getting the following error:
```Aug 17 20:13:35 master2-int prometheus[50202]: The following flag(s) can not be used in agent mode: ["--storage.tsdb.path" "--storage.tsdb.retention.time" "--storage.tsdb.retention.size"]```

File: roles/prometheus/templates/prometheus.yml.j2 [hunk 2]
```diff
@@ -17,8 +17,10 @@ remote_read:
   {{ prometheus_remote_read | to_nice_yaml(indent=2,sort_keys=False) | indent(2, False) }}
 {% endif %}
 
+{% if prometheus_rules_enable and not prometheus_agent_mode %}
```
rooty0, line 20: `rule_files` are not allowed in agent mode
```Aug 17 20:25:51 master2-int prometheus[50413]: ts=2023-08-17T20:25:51.526Z caller=main.go:479 level=error msg="Error loading config (--config.file=/etc/prometheus/prometheus.yml)" file=/etc/prometheus/prometheus.yml err="field rule_files is not allowed in agent mode"```

File: roles/prometheus/templates/prometheus.service.j2 [hunk 3]
```diff
@@ -54,6 +58,7 @@ RestrictSUIDSGID=true
 
 {% if (ansible_facts.packages.systemd | first).version is version('231', '>=') %}
 ReadWritePaths={{ prometheus_db_dir }}
+WorkingDirectory={{ prometheus_db_dir }}
```
rooty0, line 61: when `--storage.tsdb.path` is removed, prometheus is trying to write some tsdb related file to `/` aka root of filesystem, and since by default it's running from `prometheus` use we're getting an error:
```Aug 17 20:26:44 druid-master2-int prometheus[50516]: ts=2023-08-17T20:26:44.773Z caller=main.go:1155 level=error err="opening storage failed: lock DB directory: mkdir /data-agent: read-only file system"```
So defining the `WorkingDirectory` fixes this issue

Comment (SuperQ) [reply roles/prometheus/templates/prometheus.service.j2:61]: You need to set `--storage.agent.path`.

File: roles/prometheus/templates/prometheus.service.j2 (hunk 1 above)
Comment (SuperQ) [reply roles/prometheus/templates/prometheus.service.j2:16]: The flags are different for agent mode. `--storage.agent.path`, `--storage.agent.retention.min-time`, `--storage.agent.retention.max-time`.

File: roles/prometheus/templates/prometheus.service.j2 (hunk 3 above)
Comment (rooty0) [reply roles/prometheus/templates/prometheus.service.j2]: thanks, replace with `--storage.agent.path={{ prometheus_db_dir }}`

File: roles/prometheus/templates/prometheus.service.j2 (hunk 1 above)
Comment (rooty0) [reply roles/prometheus/templates/prometheus.service.j2:16]: cool, now `--storage.agent.path={{ prometheus_db_dir }}` is in place. As for `--storage.agent.retention.min-time` and `--storage.agent.retention.max-time` I guess people could use `prometheus_config_flags_extra`

Comment (SuperQ) [reply roles/prometheus/templates/prometheus.service.j2:16]: We could do a var in `vars/main.yaml` that is something like this:

```jinja
_prometheus_storage_mode: "{{ "agent" if prometheus_agent_mode else "tsdb" }}"
```

Then the flag template would be:
```jinja
  --storage.{{ _prometheus_storage_mode }}.path={{ prometheus_db_dir }}
```

rooty0: @SuperQ I'm trying to understand what the benefit we get from that? I mean we still need to
1. Include `--enable-feature=agent \` to the template
2. Exclude (remove) the flags `--storage.tsdb.retention.time` and `--storage.tsdb.retention.size`
So that being said, we still need to define `if` statement with `prometheus_agent_mode` in the template regardless of using either `--storage.{{ _prometheus_storage_mode }}.path` or `--storage.agent.path`/`--storage.tsdb.path`

SuperQ: `--storage.{{ _prometheus_storage_mode }}.retention.time`

rooty0: unfortunately not going to work
```
Aug 20 05:58:07 ip-10-16-147-81 prometheus[8786]: Error parsing command line arguments: unknown long flag '--storage.agent.retention.time'
Aug 20 05:58:07 ip-10-16-147-81 prometheus[8786]: prometheus: error: unknown long flag '--storage.agent.retention.time'
```

SuperQ: Oh, right, I wasn't looking at the agent flags closely enough.

SuperQ (approved):

gardar (commented):
Can you please add a molecule scenario for testing the agent mode?

rooty0:
> Can you please add a molecule scenario for testing the agent mode?

@gardar so just to be sure, currently there're 3 targets: `alternative`, `default` and `latest`. Do you want me to create one more target something like `agentmode` ?

gardar:
> > Can you please add a molecule scenario for testing the agent mode?
> 
> @gardar so just to be sure, currently there're 3 targets: `alternative`, `default` and `latest`. Do you want me to create one more target something like `agentmode` ?

Yes exactly!

rooty0:
hey guys, quick question
when I run default tests locally like 
- `molecule -c "~/config.yml" converge`
- `molecule -c "~/config.yml" verify`
I'm getting an error like
```
==================================== ERRORS ====================================
____________ ERROR at setup of test_version[ansible://ubuntu-22.04] ____________

    @pytest.fixture()
    def AnsibleDefaults():
>       with open("defaults/main.yml", 'r') as stream:
E       FileNotFoundError: [Errno 2] No such file or directory: 'defaults/main.yml'

tests/test_default.py:15: FileNotFoundError
=========================== short test summary info ============================
ERROR tests/test_default.py::test_version[ansible:/ubuntu-22.04] - FileNotFou...
========================= 17 passed, 1 error in 3.81s ==========================
```
the missing file is actually there in place
`# ls -la defaults/main.yml
-rw-r--r--  1 stan  staff  11223 Aug 19 00:16 defaults/main.yml`

will appreciate any help

rooty0:
@gardar So as a temporary fix I did this locally _(no commit)_ and it did the trick
```
-     with open("defaults/main.yml", 'r') as stream:
+     with open("/../../defaults/main.yml", 'r') as stream:
```

As for the tests that are running on github, I'm not sure what's going on here
1. `chrony_exporter-latest-stable-2.9` is failing but I haven't touch `chrony_exporter` at all
2. When I run `agenmode` tests **locally** 
```bash
molecule -c "/blah/ansible/ansible_collections/prometheusdev/prom/.config/molecule/config.yml" converge -s "agentmode"
``` 
All tests a passing out but:
- `centos-7` gives me an error: `FAILED! => {"changed": false, "msg": "failure 1 during daemon-reload: Failed to get D-Bus connection: No such file or directory\n"}` which looks like the image itself is messed up
- `debian-10` just crashing my `Docker Desktop` / `Virtualization framework` on my macbook **m1** (prob related) => 
```
com.apple.Virtualization.VirtualMachine crashed: {"app_name":"com.apple.Virtualization.VirtualMachine" ...
```
other platforms all passing locally

why github tests are failing not sure

need help

SuperQ:
We recently refactored our test infra. I would recommend rebasing your fork on latest main to pick up the latest changes.

roles/prometheus/molecule/agentmode/converge.yml
```diff
@@ -0,0 +1,8 @@
+---
```
SuperQ, line 1: We dropped this file and now control it centrally.

roles/prometheus/molecule/agentmode/molecule.yml
```diff
@@ -0,0 +1 @@
+---
```
SuperQ, line 1:
```suggestion
---
provisioner:
  inventory:
    group_vars:
      all:
        prometheus_agent_mode: true
```

rooty0:
@SuperQ thanks!

@gardar looks good now

SuperQ (approved):

roles/prometheus/templates/prometheus.yml.j2
gardar: You probably want to skip copying the rules when using agent mode and not just configure the service to ignore the rules.

https://github.com/prometheus-community/ansible/blob/93ed29132cd8ff042f7ba612eb8cd1eb6e20991f/roles/prometheus/tasks/configure.yml#L1-L25

Also I feel that the addition of the `prometheus_rules_enable` variable is unnecessary.
I'm unsure if we need to be able to disable the rules in setups other than agent mode.
But if we want to offer that option a better approach would be to just leave the `prometheus_alert_rules` and `prometheus_alert_rules_files` variables empty/unconfigured and disable the rules based on that.

gardar (changes requested):

rooty0:
@gardar fixed

SuperQ (approved):

gardar (approved):

PR #201, opened 2023-08-22
minor: support debian 12

gardar:


SuperQ (approved):

SuperQ:
```
Failed to find required executable \"rsync\" in paths
```

gardar:
> ```
> Failed to find required executable \"rsync\" in paths
> ```

Yep, the debian 12 container doesn't come with python so I configured molecule to prepare the image `pre_build_image: false` which should install python, etc.

But to do that the molecule docker driver uses the `synchronize` ansible module which requires `rsync`, where IMHO the `copy` module would be sufficient.

It's a catch 22 situation and I've actually thought about doing a PR to address this in the docker driver.

SuperQ (approved):
Now for Trixie.

PR #202, opened 2023-08-22
minor: support fedora 38

gardar:


SuperQ (changes requested):
Fedora 36 is EoL, let's remove it with this change.

SuperQ (approved):
I see, you're doing 36 as a separate PR.

Issue #96, opened 2023-05-05
Unable to install snmp_node_exporter


sebbbastien:
Hi,

snmp_node_exporter install seems broken. The task `prometheus.prometheus.snmp_exporter : Download snmp_exporter binary to local folder` download the binary to a "random" file (`"dest": "/tmp/8490f100-eb94-11ea-87ba-182377c269c8"`).
The task `prometheus.prometheus.snmp_exporter : Unpack snmp_exporter binary` want to unpack file from `'/tmp/snmp_exporter-0.19.0.linux-amd64.tar.gz'`.

```
TASK [prometheus.prometheus.snmp_exporter : Download snmp_exporter binary to local folder] ****************************************************************************************************************************************************************************************************************************************************
vendredi 05 mai 2023  16:01:30 +0200 (0:00:00.050)       0:01:17.256 ********** 
vendredi 05 mai 2023  16:01:30 +0200 (0:00:00.050)       0:01:17.256 ********** 
ok: [hnld-ptls-amazonl2-prometheus-dev01 -> localhost] => {
    "attempts": 1,
    "changed": false,
    "checksum_dest": "ff333c8409a9587720d2b55c31335f2290bc46c7",
    "checksum_src": "ff333c8409a9587720d2b55c31335f2290bc46c7",
    "dest": "/tmp/8490f100-eb94-11ea-87ba-182377c269c8",
    "elapsed": 1,
    "gid": 1000,
    "group": "seb",
    "md5sum": "5bccac10fb0d258148c488c30c786eca",
    "mode": "0644",
    "owner": "seb",
    "size": 7313658,
    "src": "/home/seb/.ansible/tmp/ansible-tmp-1683295290.88605-270186-112912528244913/tmpetxjw74o",
    "state": "file",
    "status_code": 200,
    "uid": 1000,
    "url": "https://github.com/prometheus/snmp_exporter/releases/download/v0.19.0/snmp_exporter-0.19.0.linux-amd64.tar.gz"
}

MSG:

OK (7313658 bytes)

TASK [prometheus.prometheus.snmp_exporter : Unpack snmp_exporter binary] **********************************************************************************************************************************************************************************************************************************************************************
vendredi 05 mai 2023  16:01:32 +0200 (0:00:01.678)       0:01:18.935 ********** 
vendredi 05 mai 2023  16:01:32 +0200 (0:00:01.679)       0:01:18.935 ********** 
[...]

PR #204, opened 2023-08-22
fix: Make binary installs consistent

SuperQ:
Update the blackbox_exporter and snmp_exporter installs to be consistent with other exporter package installs.

Fixes: https://github.com/prometheus-community/ansible/issues/96

### DIFF
--- a/roles/blackbox_exporter/tasks/install.yml
+++ b/roles/blackbox_exporter/tasks/install.yml
@@ -15,33 +15,56 @@
     createhome: false
   when: blackbox_exporter_user != 'root'
 
-- name: Download blackbox exporter binary to local folder
-  become: false
-  ansible.builtin.unarchive:
-    src: "{{ blackbox_exporter_binary_url }}"
-    dest: "/tmp"
-    remote_src: true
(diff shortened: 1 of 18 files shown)


roles/blackbox_exporter/tasks/install.yml
```diff
@@ -15,33 +15,56 @@
...
-    creates: "/tmp/blackbox_exporter-{{ blackbox_exporter_version }}.linux-{{ go_arch_map[ansible_architecture] | default(ansible_architecture) }}/\
-              blackbox_exporter"
-  register: _download_binary
-  until: _download_binary is succeeded
-  retries: 5
-  delay: 2
-  delegate_to: localhost
-  check_mode: false
-  when: not blackbox_exporter_skip_install
+- name: Discover latest version
```
gardar, line 18: I see it's also missing the ability to use the latest version lookup that the other roles have, perhaps add that while we are at it?

SuperQ: Yea, I also noticed this task is named incorrectly in all the roles.

gardar (changes requested):
Looks good, only thing missing is a molecule scenario for latest version of snmp_exporter

SuperQ:
Whups, I forgot to push that change.

PR #205, opened 2023-08-22
feat(systemd_exporter): Add TLS configuration

haavard:
This adds the systemd_exporter_tls_server_config variable, equivalent to the one in node_exporter and others, enabling the use of TLS for the systemd exporter.


roles/systemd_exporter/templates/config.yaml.j2
```diff
@@ -0,0 +1,6 @@
+---
+{{ ansible_managed | comment }}
+{% if node_exporter_tls_server_config | length > 0 %}
+tls_server_config:
+{{ node_exporter_tls_server_config | to_nice_yaml | indent(2, true) }}
```
gardar, line 5:
```suggestion
{% if systemd_exporter_tls_server_config | length > 0 %}
tls_server_config:
{{ systemd_exporter_tls_server_config | to_nice_yaml | indent(2, true) }}
```

haavard: Fixed! Cheers

gardar:
I see that you fixed the tests that were failing due to the `web.config.file` option being missing on the older version.
I think we should make the `web.config.file` flag vary depending on what version is being used, we do something similar for node_exporter
https://github.com/prometheus-community/ansible/blob/de2127bab88bc9c11c318ef3853a9e5a3333c400/roles/node_exporter/templates/node_exporter.service.j2#L26-L31

Perhaps @SuperQ knows what version of `systemd_exporter` introduced `web.config.file` ?

haavard:
@gardar it seems like systemd_exporter 0.4.0 doesn't support `--web.config` either:

    systemd_exporter: error: unknown long flag '--web.config', try --help

I added a preflight check instead to ensure we don't silently ignore the TLS config if the version is too low, let me know what you think.

SuperQ (approved):

gardar (approved):

PR #208, opened 2023-08-26
fix(systemd_exporter): Fix collector flags for older versions

haavard:
In systemd_exporter 0.5.0, `--systemd.collector` were renamed from `--collector` to `--systemd.collector`.

### DIFF
--- a/roles/systemd_exporter/molecule/alternative/molecule.yml
+++ b/roles/systemd_exporter/molecule/alternative/molecule.yml
@@ -7,3 +7,6 @@ provisioner:
         systemd_exporter_web_listen_address: "127.0.0.1:9000"
         go_arch: amd64
         systemd_exporter_version: 0.4.0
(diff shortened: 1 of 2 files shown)


SuperQ (changes requested):
Would you please add some molecule tests?

You can see `roles/prometheus/molecule/alternative/tests/test_alternative.py` for a good example of how to do this.

haavard:
> Would you please add some molecule tests?
> 
> You can see `roles/prometheus/molecule/alternative/tests/test_alternative.py` for a good example of how to do this.

@SuperQ I added these variables to the existing alternative test, which runs on the lower 0.4.0 version of systemd_exporter. The test would fail without the accompanying patch.

SuperQ:
Yea, I suppose that's ok.

SuperQ (approved):

gardar (approved):

PR #213, opened 2023-08-30
enhancement: allows using multiple web listen addresses

### LINKED ISSUE #115: [alertmanager] support multiple addresses with web.listen-address in service
The alertmanager binary supports mutliple listen addresses, from `--help`:

```
      --web.listen-address=:9093 ...
                                 Addresses on which to expose metrics and web 
                                 interface. Repeatable for multiple addresses.
```

The current service template and variable does not seem to support this. I edited manually to check what it would need to look like: 

```
ExecStart=/usr/local/bin/alertmanager \
  ...
  --web.listen-address=10.10.20.23:9093 \
  --web.listen-address=127.0.1.1:9093 \
  ...
```

```
$ netstat -nlptu | rg alertmanager | column -t
tcp  0  0  127.0.1.1:9093    0.0.0.0:*  LISTEN  105724/alertmanager
tcp  0  0  10.10.20.23:9093  0.0.0.0:*  LISTEN  105724/alertmanager
```

I guess for this change to be backwards compatible, the `alertmanager_web_listen_address` variable needs to be checked if it's either a single value or a list.

wookietreiber:
these variants should work now:

```yml
alertmanager_web_listen_address: '127.0.0.1:9093'

alertmanager_web_listen_address:
  - '127.0.0.1:9093'
  - '127.0.1.1:9093'
```

---

fixes #115

### DIFF
--- a/roles/blackbox_exporter/templates/blackbox_exporter.service.j2
+++ b/roles/blackbox_exporter/templates/blackbox_exporter.service.j2
@@ -16,7 +16,16 @@ ExecStart=/usr/local/bin/blackbox_exporter \
   {% for flag, flag_value in blackbox_exporter_cli_flags.items() -%}
   --{{ flag }}={{ flag_value }} \
   {% endfor -%}
+{% if blackbox_exporter_version is version('0.23.0', '>=') and
+      blackbox_exporter_web_listen_address is iterable and
+      blackbox_exporter_web_listen_address is not mapping and
+      blackbox_exporter_web_listen_address is not string %}
+{%   for address in blackbox_exporter_web_listen_address %}
+  --web.listen-address={{ address }}{{ " \\" if not loop.last else "" }}
+{%   endfor %}
+{% else %}
   --web.listen-address={{ blackbox_exporter_web_listen_address }}
+{% endif %}
 
 SyslogIdentifier=blackbox_exporter
 KillMode=process
--- a/roles/chrony_exporter/molecule/alternative/tests/test_alternative.py
+++ b/roles/chrony_exporter/molecule/alternative/tests/test_alternative.py
@@ -3,6 +3,7 @@ __metaclass__ = type
 
 import os
 import testinfra.utils.ansible_runner
+import pytest
 
 testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
     os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
@@ -37,10 +38,9 @@ def test_protecthome_property(host):
     assert p.get("ProtectHome") == "yes"
 
 
-def test_socket(host):
-    sockets = [
-        "tcp://127.0.0.1:8080"
-    ]
-    for socket in sockets:
-        s = host.socket(socket)
-        assert s.is_listening
+@pytest.mark.parametrize("sockets", [
+    "tcp://127.0.0.1:8080",
+    "tcp://127.0.1.1:8080",
+])
+def test_socket(host, sockets):
+    assert host.socket(sockets).is_listening
--- a/roles/alertmanager/molecule/alternative/molecule.yml
+++ b/roles/alertmanager/molecule/alternative/molecule.yml
@@ -6,7 +6,9 @@ provisioner:
         alertmanager_binary_local_dir: '/tmp/alertmanager-linux-amd64'
         alertmanager_config_dir: /opt/am/etc
         alertmanager_db_dir: /opt/am/lib
-        alertmanager_web_listen_address: '127.0.0.1:9093'
+        alertmanager_web_listen_address:
+          - '127.0.0.1:9093'
+          - '127.0.1.1:9093'
         alertmanager_web_external_url: 'http://localhost:9093/alertmanager'
         alertmanager_resolve_timeout: 10m
         alertmanager_slack_api_url: "http://example.com"
@@ -31,4 +33,4 @@ provisioner:
           peers:
             - "127.0.0.1:6783"
             - "alertmanager.demo.do.prometheus.io:6783"
-        version: 0.19.0
+        alertmanager_version: 0.25.0
--- a/roles/alertmanager/molecule/alternative/prepare.yml
+++ b/roles/alertmanager/molecule/alternative/prepare.yml
@@ -6,9 +6,9 @@
     - name: Download alertmanager binary to local folder
       become: false
       ansible.builtin.get_url:
-        url: "https://github.com/prometheus/alertmanager/releases/download/v{{ version\
-          \ }}/alertmanager-{{ version }}.linux-amd64.tar.gz"
-        dest: "/tmp/alertmanager-{{ version }}.linux-amd64.tar.gz"
+        url: "https://github.com/prometheus/alertmanager/releases/download/v{{ alertmanager_version\
+          \ }}/alertmanager-{{ alertmanager_version }}.linux-amd64.tar.gz"
+        dest: "/tmp/alertmanager-{{ alertmanager_version }}.linux-amd64.tar.gz"
         mode: 0644
       register: _download_archive
       until: _download_archive is succeeded
@@ -19,15 +19,15 @@
     - name: Unpack alertmanager binaries
       become: false
       ansible.builtin.unarchive:
-        src: "/tmp/alertmanager-{{ version }}.linux-amd64.tar.gz"
+        src: "/tmp/alertmanager-{{ alertmanager_version }}.linux-amd64.tar.gz"
         dest: "/tmp"
-        creates: "/tmp/alertmanager-{{ version }}.linux-amd64/alertmanager"
+        creates: "/tmp/alertmanager-{{ alertmanager_version }}.linux-amd64/alertmanager"
       check_mode: false
 
     - name: Link to alertmanager binaries directory
       become: false
       ansible.builtin.file:
-        src: "/tmp/alertmanager-{{ version }}.linux-amd64"
+        src: "/tmp/alertmanager-{{ alertmanager_version }}.linux-amd64"
         dest: "/tmp/alertmanager-linux-amd64"
         state: link
       check_mode: false
--- a/roles/alertmanager/molecule/alternative/tests/test_alternative.py
+++ b/roles/alertmanager/molecule/alternative/tests/test_alternative.py
@@ -47,7 +47,8 @@ def test_service(host):
 
 @pytest.mark.parametrize("sockets", [
     "tcp://127.0.0.1:9093",
-    "tcp://127.0.0.1:6783"
+    "tcp://127.0.1.1:9093",
+    "tcp://127.0.0.1:6783",
 ])
 def test_socket(host, sockets):
     assert host.socket(sockets).is_listening
--- a/roles/alertmanager/tasks/preflight.yml
+++ b/roles/alertmanager/tasks/preflight.yml
@@ -20,6 +20,27 @@
   ansible.builtin.package_facts:
   when: "not 'packages' in ansible_facts"
 
+- name: Assert that used version supports listen address type
+  ansible.builtin.assert:
+    that:
+      - >-
+        alertmanager_web_listen_address is string
+        or
+        (
+          alertmanager_version is version('0.25.0', '>=') and
+          alertmanager_web_listen_address | type_debug == "list"
+        )
+
+- name: Naive assertion of proper listen address
+  ansible.builtin.assert:
+    that:
+      - >-
+        [alertmanager_web_listen_address] |
+          flatten |
+          reject('match', '.+:\\d+$') |
+          list |
+          length == 0
+
 - name: Discover latest version
   ansible.builtin.set_fact:
     alertmanager_version: "{{ (lookup('url', 'https://api.github.com/repos/prometheus/alertmanager/releases/latest', headers=_github_api_headers,
--- a/roles/alertmanager/templates/alertmanager.service.j2
+++ b/roles/alertmanager/templates/alertmanager.service.j2
@@ -33,7 +33,16 @@ ExecStart={{ _alertmanager_binary_install_dir }}/alertmanager \
 {% endfor %}
   {{ pre }}-config.file={{ alertmanager_config_dir }}/alertmanager.yml \
   {{ pre }}-storage.path={{ alertmanager_db_dir }} \
+{% if alertmanager_version is version('0.25.0', '>=') and
+      alertmanager_web_listen_address is iterable and
+      alertmanager_web_listen_address is not mapping and
+      alertmanager_web_listen_address is not string %}
+{%   for address in alertmanager_web_listen_address %}
+  {{ pre }}-web.listen-address={{ address }} \
+{%   endfor %}
+{% else %}
   {{ pre }}-web.listen-address={{ alertmanager_web_listen_address }} \
+{% endif %}
   {{ pre }}-web.external-url={{ alertmanager_web_external_url }}{% for flag, flag_value in alertmanager_config_flags_extra.items() %} \
   {{ pre }}-{{ flag }}={{ flag_value }}{% endfor %}
 
--- a/roles/blackbox_exporter/molecule/alternative/molecule.yml
+++ b/roles/blackbox_exporter/molecule/alternative/molecule.yml
@@ -3,10 +3,13 @@ provisioner:
   inventory:
     group_vars:
       all:
-        blackbox_exporter_web_listen_address: "127.0.0.1:9000"
+        blackbox_exporter_web_listen_address:
+          - '127.0.0.1:9000'
+          - '127.0.1.1:9000'
         blackbox_exporter_cli_flags:
           log.level: "warn"
         blackbox_exporter_configuration_modules:
           tcp_connect:
             prober: tcp
             timeout: 5s
+        blackbox_exporter_version: 0.23.0
--- a/roles/blackbox_exporter/molecule/alternative/tests/test_alternative.py
+++ b/roles/blackbox_exporter/molecule/alternative/tests/test_alternative.py
@@ -33,6 +33,9 @@ def test_service(host):
         raise  # Re-raise the original assertion error
 
 
-def test_socket(host):
-    s = host.socket("tcp://127.0.0.1:9000")
-    assert s.is_listening
+@pytest.mark.parametrize("sockets", [
+    "tcp://127.0.0.1:9000",
+    "tcp://127.0.1.1:9000",
+])
+def test_socket(host, sockets):
+    assert host.socket(sockets).is_listening
--- a/roles/blackbox_exporter/tasks/preflight.yml
+++ b/roles/blackbox_exporter/tasks/preflight.yml
@@ -20,10 +20,26 @@
   ansible.builtin.package_facts:
   when: "not 'packages' in ansible_facts"
 
+- name: Assert that used version supports listen address type
+  ansible.builtin.assert:
+    that:
+      - >-
+        blackbox_exporter_web_listen_address is string
+        or
+        (
+          blackbox_exporter_version is version('0.23.0', '>=') and
+          blackbox_exporter_web_listen_address | type_debug == "list"
+        )
+
 - name: Naive assertion of proper listen address
   ansible.builtin.assert:
     that:
-      - "':' in blackbox_exporter_web_listen_address"
+      - >-
+        [blackbox_exporter_web_listen_address] |
+          flatten |
+          reject('match', '.+:\\d+$') |
+          list |
+          length == 0
 
 - name: Discover latest version
   ansible.builtin.set_fact:
--- a/roles/chrony_exporter/molecule/alternative/molecule.yml
+++ b/roles/chrony_exporter/molecule/alternative/molecule.yml
@@ -4,7 +4,9 @@ provisioner:
     group_vars:
       all:
         chrony_exporter_binary_local_dir: "/tmp/chrony_exporter-linux-amd64"
-        chrony_exporter_web_listen_address: "127.0.0.1:8080"
+        chrony_exporter_web_listen_address:
+          - '127.0.0.1:8080'
+          - '127.0.1.1:8080'
         chrony_exporter_enabled_collectors:
           - sources
         chrony_exporter_disabled_collectors:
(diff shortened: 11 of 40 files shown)


gardar (commented):
Looks like the version of alertmanager used in the alternative molecule test is too old for multiple web listen addresses.
Please add a version check in the template so that we won't attempt to use the unsupported flag on older versions.

We should probably also add support for multiple web listen addresses in the other roles where possible.
https://github.com/prometheus/exporter-toolkit/issues/91

wookietreiber:
> Please add a version check in the template so that we won't attempt to use the unsupported flag on older versions.

Do you know which version this was introduced? The changelog doesn't say (searched for both listen-address and toolkit) and I really don't wanna browse `git log -p` :sweat_smile:

gardar:
> > Please add a version check in the template so that we won't attempt to use the unsupported flag on older versions.
> 
> Do you know which version this was introduced? The changelog doesn't say (searched for both listen-address and toolkit) and I really don't wanna browse `git log -p` 😅
> 
> **Edit:** Okay, so I did browse the `git log -p` of prometheus/alertmanager a bit. I went with 0.25.0 for now, although I'm not sure at all if that's really the first version supporting multiple web listen addresses.

I think the change was introduced in this PR https://github.com/prometheus/exporter-toolkit/pull/95 so any version which uses exporter-toolkit v0.8.0 or later (https://github.com/prometheus/exporter-toolkit/commit/6b1221e7939a58e8aa767b7dc6d02140b34d31e3)
You can see that in alertmanager v0.25.0 the version of exporter-toolkit was bumped to v0.8.1 https://github.com/prometheus/alertmanager/commit/ce7b4756974d6896f95b9d26466c9a6447b9435f

Can you add multiple web listen address support to the other roles as well?

wookietreiber:
> > 
> 
> I think the change was introduced in this PR [prometheus/exporter-toolkit#95](https://github.com/prometheus/exporter-toolkit/pull/95) so any version which uses exporter-toolkit v0.8.0 or later ([prometheus/exporter-toolkit@6b1221e](https://github.com/prometheus/exporter-toolkit/commit/6b1221e7939a58e8aa767b7dc6d02140b34d31e3)) You can see that in alertmanager v0.25.0 the version of exporter-toolkit was bumped to v0.8.1 [prometheus/alertmanager@ce7b475](https://github.com/prometheus/alertmanager/commit/ce7b4756974d6896f95b9d26466c9a6447b9435f)

@gardar thanks, I was looking for exporter-toolkit in this collection :sweat_smile:

wookietreiber:
> Can you add multiple web listen address support to the other roles as well?

Will do.

wookietreiber:
I will need to bump all the versions in the `*/alternative/molecule.yml` for the tests to succeed.

wookietreiber:
Note: systemd_exporter doesn't have a release since exporter-toolkit 0.8+, so I'm not touching that one.

wookietreiber:
Also gonna use `@pytest.mark.parameterize("sockets", ...)` in all the roles I touch, the inconsistency annoys me :sweat_smile:

wookietreiber:
> > Can you add multiple web listen address support to the other roles as well?
> 
> Will do.

Done, searched for `_web_listen_address:` and added the multiple web listen address to all the roles, except for the `systemd_exporter` which doesn't have a stable release since the relevant toolkit addition.

wookietreiber:
Just rebased. Is there anything standing in the way of this getting merged and released?

SuperQ (approved):

wookietreiber:
I don't know what to do about the failing CI cases, don't even know, if/how this is related to this PR. Please advise or feel free to take over.

roles/chrony_exporter/molecule/alternative/tests/test_alternative.py
```diff
@@ -37,10 +37,9 @@ def test_protecthome_property(host):
...
 
 
-def test_socket(host):
-    sockets = [
-        "tcp://127.0.0.1:8080"
-    ]
-    for socket in sockets:
-        s = host.socket(socket)
-        assert s.is_listening
+@pytest.mark.parametrize("sockets", [
```
SuperQ, line 40: You added a pytest, but forgot to `import pytest` at the top.

SuperQ (changes requested):
Looks like a few of the test files are missing `import pytest`.

```
ERROR: roles/chrony_exporter/molecule/alternative/tests/test_alternative.py:40:1: undefined-variable: Undefined variable 'pytest'
ERROR: roles/mysqld_exporter/molecule/alternative/tests/test_alternative.py:24:1: undefined-variable: Undefined variable 'pytest'
ERROR: roles/node_exporter/molecule/alternative/tests/test_alternative.py:39:1: undefined-variable: Undefined variable 'pytest'
ERROR: roles/pushgateway/molecule/alternative/tests/test_alternative.py:40:1: undefined-variable: Undefined variable 'pytest'
ERROR: roles/smokeping_prober/molecule/alternative/tests/test_alternative.py:40:1: undefined-variable: Undefined variable 'pytest'
```

roles/chrony_exporter/molecule/alternative/tests/test_alternative.py
wookietreiber: thanks, I assumed that was already imported by all the test files :facepalm:

gardar (approved):

gardar (changes requested):
Looks like a few assert tasks need to be adjusted, but otherwise LGTM

wookietreiber:
> Looks like a few assert tasks need to be adjusted, but otherwise LGTM

I'm thinking about this for the asserts:

```yaml

- name: Naive assertion of proper listen address
  ansible.builtin.assert:
    that:
      - "':' in node_exporter_web_listen_address"
  when:
    - node_exporter_web_listen_address is defined
    - node_exporter_web_listen_address is string

- name: Naive assertion of proper listen address
  ansible.builtin.assert:
    that:
      - "':' in item"
  when:
    - node_exporter_web_listen_address is defined
    - node_exporter_web_listen_address is iterable
    - node_exporter_web_listen_address is not mapping
    - node_exporter_web_listen_address is not string
  with_items: "{{ node_exporter_web_listen_address }}"
```

What do you think @gardar @SuperQ? Dunno if this could be simplified :thinking:

gardar:
> What do you think @gardar @SuperQ? Dunno if this could be simplified 🤔

How about something like:

 ```yaml
- name: Naive assertion of proper listen address
  ansible.builtin.assert:
    that: "[node_exporter_web_listen_address] | flatten | reject('match', '.+:\\d+$') | length == 0"
```

To explain:
  * `[node_exporter_web_listen_address]` turns the var into a list, no matter if it's a single string or a list already.
  * Then `flatten` makes sure we don't end up with a nested list, in case the variable was a list already.
  * Then we reject any items from the list that match the regex `.+:\d+$`
  * Then we assert that the resulting list doesn't contain any items.

**Comment by wookietreiber** [comment]
> > What do you think @gardar @SuperQ? Dunno if this could be simplified 🤔
> 
> How about something like:
> 
> ```yaml
> - name: Naive assertion of proper listen address
>   ansible.builtin.assert:
>     that: "[node_exporter_web_listen_address] | flatten | reject('match', '.+:\\d+$') | length == 0"
> ```

Thanks @gardar that's the simplification I was looking for :sweat_smile: I just pushed the updated asserts and also implemented them for the roles that don't yet themselves support multiple web listen addresses. Let's hope that the CI works out now :crossed_fingers:

**Comment by SuperQ** [comment]
Looks like the new commits are missing the DCO.

**Comment by wookietreiber** [comment]
> fatal: [ubuntu-20.04]: FAILED! => {"msg": "The conditional check '[alertmanager_web_listen_address] | flatten | reject('match', '.+:\\d+$') | length == 0' failed. The error was: Unexpected templating type error occurred on ({% if [alertmanager_web_listen_address] | flatten | reject('match', '.+:\\d+$') | length == 0 %} True {% else %} False {% endif %}): object of type 'generator' has no len()"}

I suppose this could be fixed with `reject() | list | length` as it occurs only with Ansible 2.9?

**Comment by wookietreiber** [comment]
Don't know what to do about these: 

> fatal: [fedora-37]: FAILED! => {"changed": false, "msg": "Unable to start service prometheus: Job for prometheus.service failed because the control process exited with error code.\nSee \"systemctl status prometheus.service\" and \"journalctl -xeu prometheus.service\" for details.\n"}

**Comment by gardar** [comment]
> > fatal: [ubuntu-20.04]: FAILED! => {"msg": "The conditional check '[alertmanager_web_listen_address] | flatten | reject('match', '.+:\d+$') | length == 0' failed. The error was: Unexpected templating type error occurred on ({% if [alertmanager_web_listen_address] | flatten | reject('match', '.+:\d+$') | length == 0 %} True {% else %} False {% endif %}): object of type 'generator' has no len()"}
> 
> I suppose this could be fixed with `reject() | list | length` as it occurs only with Ansible 2.9?

Yeah type casting to a list should fix the issue.



> Don't know what to do about these:
> 
> > fatal: [fedora-37]: FAILED! => {"changed": false, "msg": "Unable to start service prometheus: Job for prometheus.service failed because the control process exited with error code.\nSee "systemctl status prometheus.service" and "journalctl -xeu prometheus.service" for details.\n"}
> 
> > fatal: [debian-11]: FAILED! => {"changed": false, "msg": "Unable to reload service snmp_exporter: snmp_exporter.service is not active, cannot reload.\n"}


That one is a bit tricky to diagnose from the current logs as by default ansible doesn't print the service failure log output.
It does suggest that there is something wrong with the service template that's preventing the service from running, it could be either a syntax error or a versioning issue where the generated service config has some unsupported options in it.
Could you run the molecule test locally with debug output? As we currently don't have a easy way of catching the error in the CI.

**Comment by gardar** [comment]
Looks like it's because the molecule alternative test is using the variable `version` and not `prometheus_version`
That test is configured for `prometheus_binary_local_dir` install where the role does not download the binary but instead uses a locally provided one, so that it actually never needs the `prometheus_version` variable, until now.
I think you could safely change the `version` variable to `prometheus_version` in `molecule/alternative/prepare.yml` and `molecule/alternative/molecule.yml`

With that being said, perhaps we also want to verify that there are not multiple listeners configured when using a old version, either by doing it in a assert task or we could also just grab the first item from the listener list in the template when using a old version (probably not ideal to do it in the template - might confuse users).

**Comment by wookietreiber** [comment]
> I think you could safely change the `version` variable to `prometheus_version` in `molecule/alternative/prepare.yml` and `molecule/alternative/molecule.yml`

Done. Did the same for for the `alertmanager` role, it was also using a plain `version`.

**Comment by wookietreiber** [comment]
> With that being said, perhaps we also want to verify that there are not multiple listeners configured when using a old version, either by doing it in a assert task or we could also just grab the first item from the listener list in the template when using a old version (probably not ideal to do it in the template - might confuse users).

Yes, I'm gonna do that next. As a separate assert or can/should I add another `that:` expression to the existing `naive assertion of proper listen address`?

**Comment by gardar** [comment]
> @gardar That seems to work for the new assert:
> 
> ```yaml
> - name: Assert that used version supports listen address type
>   ansible.builtin.assert:
>     that:
>       - >-
>         prometheus_web_listen_address is string
>         or
>         (
>           prometheus_version is version('2.40.0', '>=') and
>           prometheus_web_listen_address | type_debug == "list"
>         )
> 
> - name: Naive assertion of proper listen address
>   ansible.builtin.assert:
>     that:
>       - >-
>         [prometheus_web_listen_address] |
>           flatten |
>           reject('match', '.+:\\d+$') |
>           list |
>           length == 0
> ```
> 
> Looks good?

Yes, that looks clean, I like it :)
But you could also merge it into a single task if you want to have it even more compact, since the assert module accepts a list of expressions.

**Review by gardar — APPROVED** [review]

File: roles/blackbox_exporter/templates/blackbox_exporter.service.j2 [hunk 2]
```diff
@@ -16,7 +16,16 @@ ExecStart=/usr/local/bin/blackbox_exporter \
   {% for flag, flag_value in blackbox_exporter_cli_flags.items() -%}
   --{{ flag }}={{ flag_value }} \
   {% endfor -%}
+{% if blackbox_exporter_version is version('0.23.0', '>=') and
+      blackbox_web_listen_address is iterable and
+      blackbox_web_listen_address is not mapping and
+      blackbox_web_listen_address is not string %}
+{%   for address in blackbox_web_listen_address %}
```
SuperQ, line 23:
```suggestion
      blackbox_exporter_web_listen_address is iterable and
      blackbox_exporter_web_listen_address is not mapping and
      blackbox_exporter_web_listen_address is not string %}
{%   for address in blackbox_exporter_web_listen_address %}
```

wookietreiber: Thanks, I missed that, did the fixup.

wookietreiber:
I think I fixed the smokeping prober alternative issue, was missig a `\` in the service file for the multiple web listen addresses.

wookietreiber:
For snmp exporter alternative, I've added these tasks to the end of `main.yml`:

```yml
- command: systemctl status snmp_exporter.service
  failed_when: no
  register: status

- debug: var=status.stdout_lines

- command: journalctl -xe
  failed_when: no
  register: journal

- debug: var=journal.stdout_lines
```

(I think, that's what we should do for all services in CI to get better error messages.)

```ini
[Unit]
Description=Prometheus SNMP Exporter
After=network-online.target
StartLimitInterval=0

[Service]
Type=simple
User=nobody
Group=nogroup
ExecReload=/bin/kill -HUP $MAINPID
ExecStart=/usr/local/bin/snmp_exporter \
  --web.listen-address=127.0.0.1:9116 \
  --web.listen-address=127.0.1.1:9116 \
  --log.level=info \
  --config.file=/etc/snmp_exporter/snmp.yml

KillMode=process

SyslogIdentifier=snmp_exporter
Restart=always
RestartSec=1

[Install]
WantedBy=multi-user.target
```

```console
$ systemctl status snmp_exporter.service
● snmp_exporter.service - Prometheus SNMP Exporter",
     Loaded: loaded (/etc/systemd/system/snmp_exporter.service; enabled; vendor preset: enabled)
     Active: activating (auto-restart) (Result: exit-code) since Tue 2023-11-28 10:13:06 UTC; 669ms ago
    Process: 730 ExecStart=/usr/local/bin/snmp_exporter --web.listen-address=127.0.0.1:9116 --web.listen-address=127.0.1.1:9116 --log.level=info --config.file=/etc/snmp_exporter/snmp.yml (code=exited, status=1/FAILURE)
   Main PID: 730 (code=exited, status=1/FAILURE)
        CPU: 3ms
```

```console
$ journalctl -xe
Nov 28 10:24:49 ubuntu-22.04 snmp_exporter[736]: ts=2023-11-28T10:24:49.500Z caller=main.go:148 level=info msg=\"Starting snmp_exporter\" version=\"(version=0.21.0, branch=HEAD, revision=0d8c3527cac0c26f1d6005b84b74413d14264c37)\"
Nov 28 10:24:49 ubuntu-22.04 snmp_exporter[736]: ts=2023-11-28T10:24:49.500Z caller=main.go:149 level=info build_context=\"(go=go1.19.3, user=root@51dfd4b1f59b, date=20221122-15:14:18)\"
Nov 28 10:24:49 ubuntu-22.04 snmp_exporter[736]: ts=2023-11-28T10:24:49.500Z caller=main.go:157 level=error msg=\"Error parsing config file\" err=\"yaml: unmarshal errors:\\n  line 3: field if_mib not found in type config.plain\"
Nov 28 10:24:49 ubuntu-22.04 systemd[1]: snmp_exporter.service: Main process exited, code=exited, status=1/FAILURE
```

```
yaml: unmarshal errors:
line 3: field if_mib not found in type config.plain
```

This might be due to `snmp_exporter_version: 0.21.0` which I've added to allow for multiple web addresses. The default from `defaults/main.yml` is `snmp_exporter_version: 0.24.1`. Has configuration considerably changed between these two versions? I don't know/use snmp exporter, help appreciated.

wookietreiber:
The last one I don't get working is `smokeping_prober`, succeeds locally, but fails in CI:

```console
$ ANSIBLE_DIFF_ALWAYS=True molecule test -s alternative -p ubuntu-22.04

...

--- before                                                                                                                                                                                                 [373/9543]
+++ after: /home/umcdev/.ansible/tmp/ansible-local-245143asywggq4/tmp4d3313t_/smokeping_prober.service.j2
@@ -0,0 +1,41 @@
+#
+# Ansible managed: Do NOT edit this file manually!
+#
+[Unit]
+Description=Smokeping Prober
+After=network-online.target
+StartLimitInterval=0
+StartLimitIntervalSec=0
+
+[Service]
+Type=simple
+User=smokeping
+Group=smokeping
+PermissionsStartOnly=true
+ExecReload=/bin/kill -HUP $MAINPID
+ExecStart=/usr/local/bin/smokeping_prober \
+  --config.file=/etc/smokeping_prober//probes.yml \
+  --web.listen-address=127.0.0.1:8080 \
+  --web.listen-address=127.0.1.1:8080 \
+
+SyslogIdentifier=smokeping_prober
+KillMode=process
+Restart=always
+RestartSec=5
+
+LockPersonality=true
+NoNewPrivileges=true
+MemoryDenyWriteExecute=true
+PrivateTmp=true
+ProtectHome=true
+RemoveIPC=true
+RestrictSUIDSGID=true
+
+AmbientCapabilities=CAP_NET_RAW
+ProtectControlGroups=true
+ProtectKernelModules=true
+ProtectKernelTunables=yes
+ProtectSystem=strict
+
+[Install]
+WantedBy=multi-user.target

...

INFO     Running alternative > verify
INFO     Executing Testinfra tests found in /home/umcdev/src/ansible/prometheus/roles/smokeping_prober/molecule/alternative/tests/...
============================= test session starts ==============================
platform linux -- Python 3.11.6, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/umcdev
plugins: testinfra-10.0.0
collected 5 items

tests/test_alternative.py .....                                          [100%]

============================== 5 passed in 1.79s ===============================
INFO     Verifier completed successfully.
```

SuperQ:
Yes, I recently did a major rewrite of the snmp_exporter config, so configs need to match the version.

If you want, I can open a separate PR to make that test cleaner.

wookietreiber:
blackbox exporter alternative fails, but only on debian 10:

```
[Unit]
Description=Blackbox Exporter
After=network-online.target
StartLimitInterval=0
StartLimitIntervalSec=0

[Service]
Type=simple
User=blackbox-exp
Group=blackbox-exp
PermissionsStartOnly=true
ExecReload=/bin/kill -HUP $MAINPID
ExecStart=/usr/local/bin/blackbox_exporter \
  --config.file=/etc/blackbox_exporter.yml \
  --log.level=warn \
    --web.listen-address=127.0.0.1:9000 \
  --web.listen-address=127.0.1.1:9000 \

SyslogIdentifier=blackbox_exporter
KillMode=process
Restart=always
RestartSec=5

LockPersonality=true
NoNewPrivileges=true
MemoryDenyWriteExecute=true
PrivateTmp=true
ProtectHome=true
RemoveIPC=true
RestrictSUIDSGID=true

AmbientCapabilities=CAP_NET_RAW
ProtectControlGroups=true
ProtectKernelModules=true
ProtectKernelTunables=yes
ProtectSystem=strict

[Install]
WantedBy=multi-user.target
```

        "-- Logs begin at Thu 2023-11-30 09:06:49 UTC, end at Thu 2023-11-30 09:07:35 UTC. --",
        "Nov 30 09:07:32 debian-10 systemd[1]: /etc/systemd/system/blackbox_exporter.service:22: Unknown lvalue 'RestrictSUIDSGID' in section 'Service', ignoring",
        "Nov 30 09:07:33 debian-10 systemd[1]: /etc/systemd/system/blackbox_exporter.service:22: Unknown lvalue 'RestrictSUIDSGID' in section 'Service', ignoring",
        "Nov 30 09:07:33 debian-10 systemd[1]: Started Blackbox Exporter.",
        "-- Subject: A start job for unit blackbox_exporter.service has finished successfully",
        "-- Defined-By: systemd",
        "-- Support: https://www.debian.org/support",
        "-- ",
        "-- A start job for unit blackbox_exporter.service has finished successfully.",
        "-- ",
        "-- The job identifier is 98.",
        "Nov 30 09:07:33 debian-10 blackbox_exporter[1597]: blackbox_exporter: error: unexpected SyslogIdentifier=blackbox_exporter, try --help",
        "Nov 30 09:07:33 debian-10 systemd[1]: blackbox_exporter.service: Main process exited, code=exited, status=1/FAILURE",
        "-- Subject: Unit process exited",
        "-- Defined-By: systemd",
        "-- Support: https://www.debian.org/support",
        "-- ",
        "-- An ExecStart= process belonging to unit blackbox_exporter.service has exited.",
        "-- ",
        "-- The process' exit code is 'exited' and its exit status is 1.",
        "Nov 30 09:07:33 debian-10 systemd[1]: blackbox_exporter.service: Failed with result 'exit-code'.",
        "-- Subject: Unit failed",
        "-- Defined-By: systemd",
        "-- Support: https://www.debian.org/support",
        "-- ",
        "-- The unit blackbox_exporter.service has entered the 'failed' state with result 'exit-code'.",
        "Nov 30 09:07:33 debian-10 systemd[1]: blackbox_exporter.service: Consumed 27ms CPU time.",
        "-- Subject: Resources consumed by unit runtime",
        "-- Defined-By: systemd",
        "-- Support: https://www.debian.org/support",
        "-- ",
        "-- The unit blackbox_exporter.service completed and consumed the indicated resources."

I think the problem is this: 

> Nov 30 09:07:33 debian-10 blackbox_exporter[1597]: blackbox_exporter: error: unexpected SyslogIdentifier=blackbox_exporter, try --help

I think systemd on debian 10 is too old that it can't deal with the trailing backslash like the newer versions:

```
  --web.listen-address=127.0.1.1:9000 \

SyslogIdentifier=blackbox_exporter
```

It turns it into 

```
ExecStart=/usr/local/bin/blackbox_exporter ... --web.listen-address=127.0.1.1:9000 SyslogIdentifier=blackbox_exporter
```

I'm gonna try nukeing the trailing backslash with loop last hacks.

SuperQ:
Awesome!

wookietreiber:
Thanks for the merge and the guidance, this helped a lot.

PR #218, opened 2023-09-10
minor(prometheus): Move prometheus systemd unit task to configure

haavard:
This ensures it's part of the prometheus_configure tag, and is in line with other roles in the collection.


gardar:
Obsoleted by #425

Issue #221, opened 2023-09-27
skip installation and use Debian packages instead


anarcat:
So this role can install prometheus from a binary tarball on GitHub (which I feel is a questionable practice, but that's another issue). In our environments, we prefer packages vetted by third parties, namely for auto-upgrades and so on, so we rely on Debian packages.

It's possible to completely skip the install with `prometheus_skip_install`, but then the role will crash with:

```
[prometheus.debian.net]: FAILED! => {"changed": false, "checksum": "df2496d755394a2fd11e01e3b4bd566418adc796", "cmd": "/usr/local/bin/promtool check rules /home/anarcat/.ansible/tmp/ansible-tmp-1695839321.1357365-156317-255727393866530/source", "msg": "[Errno 2] No such file or directory: b'/usr/local/bin/promtool'", "rc": 2, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

That's because promtool is in `/usr/bin/promtool` and not `/usr/local/bin/promtool` as expected...

There's all sorts of issues like this around the package. For example, the `prometheus.service` unit file is overridden, poorly IMHO:

```
--- /usr/lib/systemd/system/prometheus.service	2023-02-04 00:20:19.000000000 +0000
+++ /etc/systemd/system/prometheus.service	2023-09-27 18:46:39.470561824 +0000
@@ -1,39 +1,53 @@
+#
+# Ansible managed
+#
+
 [Unit]
-Description=Monitoring system and time series database
-Documentation=https://prometheus.io/docs/introduction/overview/ man:prometheus(1)
-After=time-sync.target
+Description=Prometheus
+After=network-online.target
+Requires=local-fs.target
+After=local-fs.target
 
 [Service]
-Restart=on-failure
+Type=simple
+Environment="GOMAXPROCS=2"
 User=prometheus
-EnvironmentFile=/etc/default/prometheus
-ExecStart=/usr/bin/prometheus $ARGS
+Group=prometheus
 ExecReload=/bin/kill -HUP $MAINPID
-TimeoutStopSec=20s
-SendSIGKILL=no
+ExecStart=/usr/bin/prometheus \
+  --storage.tsdb.path=/var/lib/prometheus \
+  --storage.tsdb.retention.time=365d \
+  --storage.tsdb.retention.size=0 \
+  --web.config.file=/etc/prometheus/web.yml \
+  --web.console.libraries=/etc/prometheus/console_libraries \
[...]

PR #222, opened 2023-09-27
feat: allow overriding _prometheus_binary_install_dir

anarcat:
This is not ideal as the variable is still prefixed with `_` but it beats the current situation where it cannot be overriden at all.

See #221

### DIFF
--- a/roles/prometheus/defaults/main.yml
+++ b/roles/prometheus/defaults/main.yml
@@ -5,6 +5,7 @@ prometheus_binary_url: "https://github.com/{{ _prometheus_repo }}/releases/downl
                         prometheus-{{ prometheus_version }}.linux-{{ go_arch }}.tar.gz"
 prometheus_checksums_url: "https://github.com/{{ _prometheus_repo }}/releases/download/v{{ prometheus_version }}/sha256sums.txt"
 prometheus_skip_install: false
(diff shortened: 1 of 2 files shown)


l00ptr:
LGTM any news from the maintainers ?

weakcamel:
Not a maintainer, just wondering: why not rename this variable to drop `_`?

It's currently internal anyway so this change wouldn't be breaking the interface.

gardar:
Obsoleted by #425

travisdowns:
@gardar obsoleted by #425 in that sense that it fixes this issue hence no need for the PR, or in the sense that the PR cannot be applied because the code changed too much?

travisdowns:
Thanks for the reply: so it is possible to combine `prometheus_skip_install=true` and `prometheus_binary_install_dir` to point to an existing prom (e.g., one installed via package managed as described in #221)?

gardar:
Yes that should work, although I just noticed that the old internal variable was still being used in the prometheus role, but I just merged a fix for that in #456 
Please note that the `prometheus_skip_install` variable was removed in favor of using tags to control the behavior of the role, there are tags for config, install, download, etc.
Also please note that the roles write a systemd service and will potentially overwrite existing services that come from deb/rpm/etc packages, even though you skip the install.

Issue #178, opened 2023-07-29
Update blackbox_exporter config


SuperQ:
The `blackbox_exporter` role has a couple of config issues.
* It puts the exporter config in `/etc` directly.
* It is missing the web config file.

Issue #64, opened 2023-03-09
Share common tasks across roles


gardar:
We should investigate if it would make sense to share some of the common task files across roles, to avoid duplicate code.

For example much of the `selinux.yml`, `install.yml`, `configure.yml` is the same across roles.

We could collect those common task files into one role and just include the tasks we need from that role

So something like this:
From: 
https://github.com/prometheus-community/ansible/blob/d389ad29e203ed24aebb02796958837bdca6d35e/roles/node_exporter/tasks/main.yml#L24-L31

to:
```yaml
 - name: SELinux 
   ansible.builtin.include_role: 
     name: prometheus.prometheus.common
     tasks_from: selinux.yml 
     apply: 
       become: true 
   when: ansible_selinux.status == "enabled" 
   tags: 
     - node_exporter_configure 
```

It's also possible to pass variables to the role when it's being imported, so we can essentially use the common role to create `functions` which we can use across other roles.
Something like:

```yaml
- name: Install
  ansible.builtin.include_role:
    name: prometheus.prometheus.common
    tasks_from: install.yml
  vars:
    binary_url: https://example.tld/prometheus.tar.gz
    user: node-exp
```

gardar:
cc @SuperQ @paulfantom

PR #224, opened 2023-09-28
feat: bunch of improvements on blackbox_exporer

anarcat:
This puts the config file in `/etc/prometheus`, allows changes to the binary name (so we can use the Debian package directly).

Closes: #178

### DIFF
--- a/roles/blackbox_exporter/defaults/main.yml
+++ b/roles/blackbox_exporter/defaults/main.yml
@@ -75,5 +75,6 @@ blackbox_exporter_configuration_modules:
 blackbox_exporter_config_dir: /etc/blackbox_exporter
 
 blackbox_exporter_binary_install_dir: "/usr/local/bin"
+blackbox_exporter_binary_name: "blackbox_exporter"
 
 blackbox_exporter_service_name: blackbox_exporter
--- a/roles/blackbox_exporter/defaults/main.yml
+++ b/roles/blackbox_exporter/defaults/main.yml
@@ -71,7 +71,7 @@ blackbox_exporter_configuration_modules:
 #      validate_answer_rrs:
 #        fail_if_matches_regexp: [test]
 
-# Where to put the blackbox_exporter.yml main configuration file
-blackbox_exporter_config_dir: /etc
+# Where to put the blackbox.yml main configuration file
+blackbox_exporter_config_dir: /etc/blackbox_exporter
 
 blackbox_exporter_binary_install_dir: "/usr/local/bin"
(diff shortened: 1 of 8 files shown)


roles/blackbox_exporter/defaults/main.yml
```diff
@@ -71,7 +71,10 @@ blackbox_exporter_configuration_modules:
 #      validate_answer_rrs:
 #        fail_if_matches_regexp: [test]
 
-# Where to put the blackbox_exporter.yml main configuration file
-blackbox_exporter_config_dir: /etc
+# Where to put the blackbox.yml main configuration file
+blackbox_exporter_config_dir: /etc/prometheus
```
SuperQ, line 75: This should be `/etc/blackbox_exporter` to match the other roles.

roles/blackbox_exporter/defaults/main.yml
```diff
@@ -71,7 +71,10 @@ blackbox_exporter_configuration_modules:
...
 
 blackbox_exporter_binary_install_dir: "/usr/local/bin"
+blackbox_exporter_binary_name: "blackbox_exporter"
```
SuperQ, line 78: I don't think we want to do this.

SuperQ (changes requested):

roles/blackbox_exporter/defaults/main.yml
anarcat: could you expand on this? what exactly do we want to not do here, and why?

roles/blackbox_exporter/defaults/main.yml
anarcat: okay, done.

roles/blackbox_exporter/defaults/main.yml
gardar: I don't see any need/benefit with having the binary name customizable, we should keep the options at minimal and avoid having options just for options sake.

anarcat: this is not options option's sake: i have modified this option for it to be compatible with the debian package....

gardar: Ah apologies. At the moment we do not support other installation methods than the binary install method that's currently in the roles.
When and if we expand that into rpm/deb/etc. packages we would definitely want the roles to install those packages, then variables such as these would most likely be populated internally in the role and not user configurable.
With that being said, such a feature would be a major change and I would first like to take care of some streamlining/deduplicating of tasks in the roles we currently have. (#64 )

anarcat: ah. well, i guess one thing i'm saying is you're almost there and you just need to use the right binary name here. i've deployed prom on a debian 12 host using this playbook and it kind of worked after the bunch of patches i had to roll here...

gardar:
Obsoleted by #425

Issue #225, opened 2023-09-28
unable to run node_exporter playbook when offline and SELinux enabled on RedHat


avdland:
**Background:**
We are using RedHat with SELinux enabled in an offline environment.
When SELinux is enabled, the node_exporter playbook wants to make sure two packages are present: python3-libselinux and python3-policycoreutils.
**Problem:**
The task to check for the presence of the packages always fails due to Yum which is unable to fetch metadata from the Yum repositories. Note that the packages are already installed so why does Yum wants to fetch repo metadata.

I've noticed that when I run "yum check-update" beforehand with internet temporarily enabled, the task does succeed afterwards when offline...

gardar:
While I agree that this is [a bug in the ansible module](https://github.com/ansible/ansible/issues/81847) I don't think that this is something we should account for, it feels strange to have a host in a air gapped/offline environment but still have yum/dnf configured to use public mirrors.

In this situation I would set up a internal yum mirror, but if that's out of the question then you could even set up a local directory mirror for yum that's on the host itself.

PR #226, opened 2023-09-28
fix: dont use Yum/Rpm task when packages are already installed

avdland:
This should fix issue #225

### DIFF
--- a/roles/node_exporter/defaults/main.yml
+++ b/roles/node_exporter/defaults/main.yml
@@ -5,6 +5,7 @@ node_exporter_binary_url: "https://github.com/{{ _node_exporter_repo }}/releases
                            node_exporter-{{ node_exporter_version }}.linux-{{ go_arch }}.tar.gz"
 node_exporter_checksums_url: "https://github.com/{{ _node_exporter_repo }}/releases/download/v{{ node_exporter_version }}/sha256sums.txt"
 node_exporter_skip_install: false
+node_exporter_offline: false
 
 node_exporter_web_listen_address: "0.0.0.0:9100"
 node_exporter_web_telemetry_path: "/metrics"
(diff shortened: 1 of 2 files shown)


SuperQ (changes requested):
This basically bypasses the real functionality of `ansible.builtin.package`. Rather than have a hacky workaround, we should probably switch to `ansible.builtin.yum` directly. This would allow tweaking the offline behavior with [controlling the metadata update flags](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html#parameter-cacheonly).

gardar:
> This basically bypasses the real functionality of `ansible.builtin.package`. Rather than have a hacky workaround, we should probably switch to `ansible.builtin.yum` directly. This would allow tweaking the offline behavior with [controlling the metadata update flags](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html#parameter-cacheonly).

Agreed, but there is no need to switch to the yum module, the package module is actually just a proxy on top of the yum module, which means you can pass any of the yum module parameters to the package module

avdland:
While testing, I had set cacheonly to true, but then it complained that the cache was empty, and the task would still fail...

avdland:
But even if there was a parameter for the package module, there is no way I can set it outside of the playbook.

avdland:
It seems that the `disablerepo` parameter does exactly what I want. I've added a new variable, `node_exporter_offline`, which then disables the repos if true.

avdland:
I do think the previous commit https://github.com/prometheus-community/ansible/commit/b41c938919a749c9adcdeba6b29f552ae6cca531 is a better solution though... we shouldnt even have to run yum commands if the package is already installed

SuperQ:
IMO this is a bug in Ansible's package module. It should not need to update caches or activate yum with `state: present`.

avdland:
created new issue: https://github.com/ansible/ansible/issues/81847

gardar:
https://github.com/prometheus-community/ansible/issues/225#issuecomment-2419535914

PR #236, opened 2023-10-21
fix: Drop support for Ansible 2.11

SuperQ:
2.11 does not work with the current Galaxy API.


SuperQ:
Ping @gardar, all PRs are failing due to Galaxy API changes.

gardar:
fixed #241

PR #248, opened 2023-11-17
skip_changelog(ci): make sure pytest and pytest-testinfra versions stay compatible

gardar:
Testinfra tests are failing on python 3.9 since testinfra v10.0.0.
https://github.com/prometheus-community/ansible/actions/runs/6906336970/job/18791174950

Seems to be caused by a old version of pytest already being installed on the ci image, this should update pytest and fix the issue.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,6 @@
 molecule
 docker
+pytest
 pytest-testinfra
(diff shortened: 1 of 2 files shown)


SuperQ:
Thanks!

PR #266, opened 2023-12-18
fix: conditional statement jinja2 templating warning

iisojunn:
A new warning occurred after updating from ansible-core 2.15.4 to 2.15.8 [WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: ...

Removed the jinja2 templating delimeters to fix the warning.


gardar (approved):
Good catch, thanks!

PR #270, opened 2024-01-03
fix(mysqld_exporter): Change condition for mysqld_exporter_host check

molpako:
Modify the condition for `mysqld_exporter_host` in the mysqld_exporter my_cnf.j2 template to address an issue where `mysqld_exporter_host: null` in roles/defaults/main.yml was not being handled correctly. The updated condition ensures the host setting logic operates properly even when `mysqld_exporter_host` is set to null.

I also think a useful approach would be to simply remove mysqld_exporter_host[port] from defaults/main.yml and not change the template condition. I would appreciate your comments on this.

### DIFF
--- a/roles/mysqld_exporter/templates/my_cnf.j2
+++ b/roles/mysqld_exporter/templates/my_cnf.j2
@@ -2,7 +2,7 @@
 [client]
 user={{ mysqld_exporter_username }}
 password={{ mysqld_exporter_password }}
-{% if mysqld_exporter_host is defined %}
+{% if mysqld_exporter_host %}
 host={{ mysqld_exporter_host }}
 port={{ mysqld_exporter_port }}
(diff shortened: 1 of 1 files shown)


gardar:
Good catch, thanks!

> I also think a useful approach would be to simply remove mysqld_exporter_host[port] from defaults/main.yml and not change the template condition. I would appreciate your comments on this.

That's up for debate and it's actually a part of a larger discussion.

It's been standard practice in ansible roles that the defaults have served the purpose of giving a overview the variables that users can override. But as we recently introduced the argument spec validation of the roles which serves the same purpose (and is actually used to generate the docs) then perhaps it's ok to remove this variable and other null value ones from the defaults.

molpako:
Thank you for merging and explaining the background.

PR #272, opened 2024-01-07
feat: Add cadvisor role

SuperQ:
Add a role to deploy https://github.com/google/cadvisor.


roles/cadvisor/defaults/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+cadvisor_version: 0.47.2
+cadvisor_binary_local_dir: ""
+cadvisor_binary_url: "https://github.com/{{ _cadvisor_repo }}/releases/download/v{{ cadvisor_version }}/\
+                           cadvisor-v{{ cadvisor_version }}-linux-{{ go_arch }}"
```
gardar, line 5:
```suggestion
cadvisor_binary_url: "https://github.com/{{ _cadvisor_repo }}/releases/download/v{{ cadvisor_version }}/\
                      cadvisor-v{{ cadvisor_version }}-linux-{{ go_arch }}"
```

roles/cadvisor/meta/argument_specs.yml
```diff
@@ -0,0 +1,49 @@
+---
+# yamllint disable rule:line-length
+argument_specs:
+  main:
+    short_description: "cAdvisor"
+    description:
+      - "Deploy L(cadvisor,https://github.com/google/cadvisor) using ansible"
+    author:
+      - "cAdvisor Authors"
```
gardar, line 9: I think this field is for the role authors

roles/cadvisor/molecule/alternative/prepare.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Run local preparation
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: Download cadvisor binary to local folder
+      become: false
+      ansible.builtin.get_url:
+        url: "https://github.com/google/cadvisor/releases/download/v{{ cadvisor_version\
+          \ }}/cadvisor-v{{ cadvisor_version }}-linux-{{ go_arch }}"
```
gardar, line 10:
```suggestion
        url: "{{ cadvisor_binary_url }}"
```

gardar (changes requested):

roles/cadvisor/meta/argument_specs.yml
SuperQ: Oh, ok.

roles/cadvisor/molecule/alternative/prepare.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Run local preparation
+  hosts: localhost
```
gardar, line 3: Perhaps we need something like

```suggestion
  hosts: localhost
  vars_files: 
    - ../defaults/main.yml
```

To make `cadvisor_binary_url` variable available

SuperQ: I put this back to the original pattern, which is what we have in all of the other role tests.

gardar: Right ok, we'll put this on the back burner for all the roles then

PR #283, opened 2024-01-29
fix: Don't reload services when restarting

SuperQ:
Avoid triggering a reload of various role services when a restart is needed. This avoids the race between restart and reload.


SuperQ:
Looks like https://github.com/ansible/ansible-lint/pull/3989 broke the ansible-lint action. :crying_cat_face:

PR #290, opened 2024-02-11
enhancement: Add systemd socket listener activation

oussamasiyagh:
This allows systemd to open the listening port and enable the exporter. it can maintain the port open even if the exporter is not running. it should also open the port at boot before a process has an opportunity of taking it. in addition, it will queue TCP requests in case the exporter goes down, attempt to start the exporter, and then forward the requests to the exporter.

### DIFF
--- a/roles/node_exporter/defaults/main.yml
+++ b/roles/node_exporter/defaults/main.yml
@@ -6,7 +6,8 @@ node_exporter_binary_url: "https://github.com/{{ _node_exporter_repo }}/releases
 node_exporter_checksums_url: "https://github.com/{{ _node_exporter_repo }}/releases/download/v{{ node_exporter_version }}/sha256sums.txt"
 node_exporter_skip_install: false
 
-node_exporter_web_listen_address: "0.0.0.0:9100"
+node_exporter_web_listen_port: 9100
+node_exporter_web_listen_address: "0.0.0.0:{{ node_exporter_web_listen_port }}"
 node_exporter_web_telemetry_path: "/metrics"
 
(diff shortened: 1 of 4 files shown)


gardar (commented):
I don't really see the need to add the socket as we are already keeping the service alive with `restart=always` and also not sure if there is any real world scenarios where you need to buffer the tcp requests while the service starts due to both the nature of the service and how lightweight it is.
With that being said I'm not particularly against adding the socket, sure it sounds good on paper but I'm just not sure if there is any real benefit from it.
I'm also not sure if using the socket should be the default behavior or if it should be a optional feature.
In any case if this were to be added it should be added to the other roles as well and it should be depending on the exporter version as it is a fairly recent feature.

Thoughts @SuperQ ?

SuperQ:
Yea, when we added systemd socket activation, I never really saw the point of it. The only real nice thing is that systemd reserves the port early in boot to avoid the port being stolen.

The TCP request queue is, IMO, not a good thing from a monitoring perspective. It gives the monitoring system a false impression that the target is up.

Either way, I agree, this can not break multiple listeners.

SuperQ (changes requested):
Tests need to pass.

oussamasiyagh:
well, I thought it's worth adding it here as in [https://github.com/prometheus/node_exporter/tree/master/examples/systemd](systemd unit example), it's all about reserving the port for the node_exporter.

oussamasiyagh:
systemd socket activation is now as option when `node_exporter_systemd_socket_activation` is true

PR #294, opened 2024-02-17
feat: Add fail2ban_exporter role

anviar:
Add new role: **fail2ban_exporter**


roles/fail2ban_exporter/vars/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+go_arch_map:
+  i386: '386'
+  x86_64: 'amd64'
+  aarch64: 'arm64'
+  armv7l: 'armv7'
+  armv6l: 'armv6'
+
+go_arch: "{{ go_arch_map[ansible_architecture] | default(ansible_architecture) }}"
+_gitlab_projectid: 24199687
```
SuperQ, line 10: I guess we need to update `.github/scripts/version_updater.sh` to handle checking for new versions from GitLab projects.

SuperQ (approved):
Not going to block this on having auto-updates. But it would be good to followup with some refactoring to handle github vs gitlab.

anviar:
This project have github mirror, but I prefer to use original source.
I will check `.github/scripts/version_updater.sh`, may be I can help with that.
Should I open another PR for that update or use this one?

SuperQ:
A new PR is fine.

PR #301, opened 2024-02-22
bugfix: use temporary directory to make compatible with multi-user controllers

mark-tomich:
In our use case, we've got an Ansible controller from which many administrators launch playbooks. The first administrator becomes the owner of the tarball in /tmp...and that's okay because it's world-readable. However, when the second administrator attempts to extract the archive to /tmp, it fails because the first administrator owns the extracted location.  This patch fixes the problem...at the (small) expense of having to extract the tarball multiple times.

### DIFF
--- a/roles/smokeping_prober/defaults/main.yml
+++ b/roles/smokeping_prober/defaults/main.yml
@@ -33,3 +33,6 @@ smokeping_prober_binary_install_dir: "/usr/local/bin"
 smokeping_prober_config_dir: "/etc/smokeping_prober"
 smokeping_prober_system_group: "smokeping"
 smokeping_prober_system_user: "{{ smokeping_prober_system_group }}"
+
+# Local path to stash the archive and its extraction
+smokeping_archive_path: /tmp


gardar (commented):
I like the idea of using the `tempfile` module rather than having `/tmp` hardcoded but I have some concerns.

Since the `tempfile` module generates directories with unique, random names, doesn't this mean that for each targeted host, the binary will be extracted into a new temporary directory? This could lead to increased CPU usage due to repeatedly processing the archive and also consume more disk space than necessary. Additionally, for subsequent executions, the extraction process would have to be repeated, given that the names of temporary directories created by the tempfile module do not persist across Ansible runs.

Maybe a more efficient approach would be to simply designate a variable for the local path. This change could also bring the advantage of enabling the storage of binaries on the localhost's persistent storage.

mark-tomich:
> Since the tempfile module generates directories with unique, random names, doesn't this mean that for each targeted host, the binary will be extracted into a new temporary directory? This could lead to increased CPU usage due to repeatedly processing the archive and also consume more disk space than necessary. Additionally, for subsequent executions, the extraction process would have to be repeated, given that the names of temporary directories created by the tempfile module do not persist across Ansible runs.

Yes, you are correct on all counts. However, as it stands, it's unusable by more than one admin on a single system. I'd argue that working inefficiently is preferable to not working at all.  Also, the tarballs aren't that big to begin with, so extracting them multiple times isn't so bad...and they're cleaned up at the end in any case (unlike today).

> Maybe a more efficient approach would be to simply designate a variable for the local path. This change could also bring the advantage of enabling the storage of binaries on the localhost's persistent storage.

I could live with that, but then this PR becomes a feature rather than a bug :)

...one other concern I had with this code: as it stands, any system user could drop a trojan tarball into `/tmp` and that tarball would be dutifully extracted and propagated by anybody using this code...there is no signature validation or checksumming to prevent it. In my case, only trusted admins are permitted on the system, but if I were running in a more broadly shared environment, this would be a risk.

mark-tomich:
Alright...I've changed the approach by using in an optional `prometheus_archive_path` variable. Let me know what you think.

gardar (changes requested):
Can you please use variables that have the role name as prefix such as `alertmanager_archive_path`, `node_exporter_archive_path`, etc. rather than using `prometheus_archive_path` for all the roles.

Also please put the variables in `defaults/main.yml` with `/tmp` as the default value instead of using the `| default('/tmp')` filter. For users it's hard to track down which options you can use with the role if you have to go and read through the code to find the variables.
And last but not least please add the variables to the `meta/argument_specs.yml` which is used to validate the role options and generate the docs for the collection under https://prometheus-community.github.io/ansible/

gardar (changes requested):
Please add the variables to `meta/argument_specs.yml` as well

PR #308, opened 2024-02-27
skip_changelog(ci): gitlab.com support for version updater

anviar:
Support gitlab.com projects for `version_updater.sh`
@SuperQ as we discussed here #294 
Please, check this code, because I don't know how to test it myself completely.
My main idea was to keep code structure unchanged so much as possible.

### DIFF
--- a/.github/scripts/version_updater.sh
+++ b/.github/scripts/version_updater.sh
@@ -68,10 +68,11 @@ if [[ -z "${role}" ]]; then
 fi
 
 # Get latest version.
-if [[ ${type} == "github" ]]
+if [[ "${type}" == "github" ]]
 then
   version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
-elif [[ ${type} == "gitlab" ]]
+elif [[ "${type}" == "gitlab" ]]
+then
   version="$(curl https://gitlab.com/api/v4/projects/${source_repo}/releases|jq '.[0].tag_name'| tr -d '"v')"
 else
   echo_red 'Unknown source type. Terminating.'
--- a/.github/scripts/version_updater.sh
+++ b/.github/scripts/version_updater.sh
@@ -73,7 +73,7 @@ then
   version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
 elif [[ "${type}" == "gitlab" ]]
 then
-  version="$(curl https://gitlab.com/api/v4/projects/${source_repo}/releases|jq '.[0].tag_name'| tr -d '"v')"
+  version="$(curl --retry 5 --silent --fail "https://gitlab.com/api/v4/projects/${source_repo}/releases"|jq '.[0].tag_name'| tr -d '"v')"
 else
   echo_red 'Unknown source type. Terminating.'
   exit 128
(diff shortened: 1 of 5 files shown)


.github/scripts/version_updater.sh
```diff
@@ -67,7 +68,16 @@ if [[ -z "${role}" ]]; then
...
+if [[ "${type}" == "github" ]]
+then
+  version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
+elif [[ "${type}" == "gitlab" ]]
+then
+  version="$(curl --retry 5 --silent --fail "https://gitlab.com/api/v4/projects/${source_repo}/releases"|jq '.[0].tag_name'| tr -d '"v')"
+else
+  echo_red 'Unknown source type. Terminating.'
+  exit 128
+fi
```
gardar, line 80: Not sure if this is ever executed, as you are setting github as the default repo_type

.github/scripts/version_updater.sh
```diff
@@ -67,7 +68,16 @@ if [[ -z "${role}" ]]; then
 fi
 
 # Get latest version.
-version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
+if [[ "${type}" == "github" ]]
+then
+  version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
+elif [[ "${type}" == "gitlab" ]]
+then
+  version="$(curl --retry 5 --silent --fail "https://gitlab.com/api/v4/projects/${source_repo}/releases"|jq '.[0].tag_name'| tr -d '"v')"
```
gardar, line 76: Perhaps move the gitlab api url to a function named gitlab_api() to have it in line with the existing github_api() function.

.github/scripts/version_updater.sh
anviar: You right, but in future something can be changed and that condition can handle such situation.

.github/scripts/version_updater.sh
anviar: I thought about it. I thought this would be an overcomplication of the code.
Okay, I pushed update )

gardar (approved):

.github/scripts/version_updater.sh
```diff
@@ -67,7 +75,16 @@ if [[ -z "${role}" ]]; then
...
+if [[ "${type}" == "github" ]]
+then
+  version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
+elif [[ "${type}" == "gitlab" ]]
+then
+  version="$(gitlab_api "projects/${source_repo}/releases" | jq '.[0].tag_name' | tr -d '"v')"
+else
+  echo_red 'Unknown source type. Terminating.'
+  exit 128
+fi
```
SuperQ, line 87: Let's use a case statement here. Looks cleaner than a bunch of ifs.

```suggestion
case "${type}" in
  github)
    version="$(github_api "repos/${source_repo}/releases/latest" | jq '.tag_name' | tr -d '"v')"
    ;;
  gitlab)
    version="$(gitlab_api "projects/${source_repo}/releases" | jq '.[0].tag_name' | tr -d '"v')"
    ;;
  *)
    echo_red 'Unknown source type. Terminating.'
    exit 128
    ;;
esac
```

SuperQ (approved):

PR #323, opened 2024-03-23
enhancement(node_exporter): use systemd to create node_exporter_textfile_dir if it doesn't exist

ym:
We're currently using a directory inside tmpfs for textfile collectors to reduce possible unnecessary IO usage, and the directory will be removed after a reboot. This PR is to ensure that `node_exporter_textfile_dir` exists when node-exporter is started.


roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -8,6 +8,16 @@ After=network-online.target
 Type=simple
 User={{ node_exporter_system_user }}
 Group={{ node_exporter_system_group }}
+{% if node_exporter_textfile_dir | length > 0 %}
+{% if node_exporter_systemd_version | int >= 235 %}
```
SuperQ, line 12:
```suggestion
{% if (ansible_facts.packages.systemd | first).version is version('235', '>=') %}
```

SuperQ (changes requested):

roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -8,6 +8,16 @@ After=network-online.target
...
+ExecStartPre=+/bin/mkdir -p {{ node_exporter_textfile_dir }}
+ExecStartPre=+/bin/chown -R {{ node_exporter_system_user }}:{{ node_exporter_system_group }} {{ node_exporter_textfile_dir }}
+{% else %}
+ExecStartPre=/bin/mkdir -p {{ node_exporter_textfile_dir }}
+ExecStartPre=/bin/chown -R {{ node_exporter_system_user }}:{{ node_exporter_system_group }} {{ node_exporter_textfile_dir }}
```
SuperQ, line 17: If you move this outside the version check block, you don't have to repeat it.

roles/node_exporter/templates/node_exporter.service.j2
ym: Great point :-) updated as requested.

roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -8,6 +8,16 @@ After=network-online.target
 Type=simple
 User={{ node_exporter_system_user }}
 Group={{ node_exporter_system_group }}
+{% if node_exporter_textfile_dir | length > 0 %}
+{% if (ansible_facts.packages.systemd | first).version is version('235', '>=') %}
+ExecStartPre=+/bin/mkdir -p {{ node_exporter_textfile_dir }}
+ExecStartPre=+/bin/chown -R {{ node_exporter_system_user }}:{{ node_exporter_system_group }} {{ node_exporter_textfile_dir }}
```
SuperQ, line 14: Hmm, I wonder if this should be `-R`. I'm not sure we want all of the prom files changed, just the dir.

ym: In our environment, the text file exporters are started after `node_exporter.service` and the directory is used exclusively for text file exporters, so it doesn't really matter, but I think we can remove the `-R`, or should we go further and only change the owner if the directory is empty?

gardar:
I wonder if using systemd tmpfiles in this scenario would be a better solution? https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html

gardar (commented):
I wonder if using systemd tmpfiles in this scenario would be a better solution? https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html

gardar:
> I wonder if using systemd tmpfiles in this scenario would be a better solution? https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html

ping @ym

ym:
> > I wonder if using systemd tmpfiles in this scenario would be a better solution? https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html
> 
> ping @ym

External scripts will also need to write to `node_exporter_textfile_dir`, so I think a predictable and static path might be a better idea. But if you have any good suggestions, I'm all ears.

gardar:
The path is not dynamic like what you get with `mktemp`, you define the path you wish to be created in a config file under tmpfiles.d and then systemd will take care of creating it.

roles/node_exporter/templates/node_exporter.service.j2
```diff
@@ -8,6 +8,16 @@ After=network-online.target
 Type=simple
 User={{ node_exporter_system_user }}
 Group={{ node_exporter_system_group }}
+{% if node_exporter_textfile_dir | length > 0 %}
```
zhan9san, line 11: It would affect the common user who does not use tmpfs.

PR #330, opened 2024-04-09
skip_changelog: fix changelog screwup

gardar:


SuperQ:
Sad_Log

gardar:
No idea why this keeps happening, could it be some race condition when multiple PR's are merged at the same time?

SuperQ:
Yea, that's my guess. If you merge while it's creating a change, it fails to catch the intermediate ones.

PR #354, opened 2024-04-27
fix: Improve apt facts install

SuperQ:
Reduce the number of packages installed by the apt package facts preflight by using the apt module and disabling recommends packages.

### DIFF
--- a/roles/alertmanager/tasks/preflight.yml
+++ b/roles/alertmanager/tasks/preflight.yml
@@ -9,6 +9,7 @@
   ansible.builtin.package:
     name: "{{ _pkg_fact_req }}"
     state: present
+    install_recommends: false
   when: (_pkg_fact_req)
   vars:
     _pkg_fact_req: "{% if (ansible_pkg_mgr == 'apt') %}\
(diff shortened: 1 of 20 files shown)


gardar (commented):
There's no need to switch the task over to `apt` as the `package` module is actually just a proxy for apt or whatever package manager is present.

With that being said, the way the task is configured right now it actually never runs on any non apt systems.
I think we should make sure the task also installs the required dependencies for other distros, even though we don't run tests against them. it looks like suse requires python3-rpm and gentoo needs qlist.

SuperQ:
Except we do need to switch to `apt`, since `package` doesn't expose "install_recommends" as a parameter.

gardar:
That is true but it simply passes the parameters down to the underlying module that it's proxy-ing.
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html#id1

gardar (commented):
Might be good to only pass the `install_recommends` parameter if `ansible_pkg_mgr` is `apt` else omit it to avoid future issues.

PR #363, opened 2024-05-18
fix: Downgrade name[casing] to warning

SuperQ:
Fix linting errors by downgrading the `name[casing]` check to warning.


gardar (approved):
Agreed that this is the best course of action until this gets resolved upstream

PR #367, opened 2024-05-24
fix: make sure antsibull-changelog doesn't offend ansible-lint

gardar:
Finally figured out why the format of `changelog.yaml` gets messed up from time to time.
I thought `antsichaut` was the culpit but it turns out that `antsibull-changelog` was responsible.

The workflow first passes the changelog to `antsibull-changelog` and then to `antsichaut`.
`antsibull-changelog` rewrites the changelog even if there is no new change but `antsichaut` skips processing it when there are no new changes.
And apparently the yaml that `antsibull-changelog` was putting out was not compatible with `ansible-lint`.

I've proposed fixes over at https://github.com/ansible-community/antsibull-changelog/pull/160 and https://github.com/ansible-community/antsibull-changelog/pull/165

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
           fetch-depth: 0
 
       - name: "Install tools"
-        run: "python -m pip install ansible-base antsibull-changelog --disable-pip-version-check"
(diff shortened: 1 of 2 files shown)


SuperQ:
Looks like the upstream change has been approved. Let's give it a bit to see if they merge.

PR #292, opened 2024-02-15
Added prometheus scrape_config_files support

Lycowolf:
feat: Add support for scrape_config_files (fixes https://github.com/prometheus-community/ansible/issues/145)

PR #368, opened 2024-05-28
feat: Added prometheus scrape_config_files support

### LINKED ISSUE #145: [prometheus] - New scrape_config_files
Starting from version 2.43.0 of Prometheus, we have a new section available in the configuration file prometheus.yml called 'scrape_config_files,' where you can specify the paths of files containing the scrape_config settings. 

Would it be possible to implement this functionality by adding the 'scrape_config_files' attribute to the prometheus.yml.j2 template if it is defined in a variable, for example, 'prometheus_scrape_config_files'? It could be a list of paths."

manzsolutions-lpr:
Add support for scrape_config_files (fixes #145)

Attention: This is merely a rebase of #292.

### DIFF
--- a/roles/prometheus/meta/argument_specs.yml
+++ b/roles/prometheus/meta/argument_specs.yml
@@ -150,6 +150,14 @@ argument_specs:
         default:
           - "prometheus/targets/*.yml"
           - "prometheus/targets/*.json"
+      prometheus_scrape_config_files:
+        description:
+          - "List of folders where ansible will look for files containing custom scrape config configuration files which will be copied to C({{ prometheus_config_dir }}/scrapes/)."
+        type: "list"
+        elements: "str"
+        default:
+          - "prometheus/scrapes/*.yml"
+          - "prometheus/scrapes/*.json"
       prometheus_system_group:
         description:
           - "System group for Prometheus."
--- a/roles/prometheus/meta/argument_specs.yml
+++ b/roles/prometheus/meta/argument_specs.yml
@@ -153,6 +153,7 @@ argument_specs:
       prometheus_scrape_config_files:
         description:
           - "List of folders where ansible will look for files containing custom scrape config configuration files which will be copied to C({{ prometheus_config_dir }}/scrapes/)."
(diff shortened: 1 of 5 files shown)


gardar (changes requested):

gardar (changes requested):
We need to make sure the `scrape_config_files` option is not configured for older versions that don't support the option.

roles/prometheus/meta/argument_specs.yml
```diff
@@ -112,6 +112,7 @@ argument_specs:
       prometheus_scrape_configs:
         description:
           - "Prometheus scrape jobs provided in same format as in the L(official docs,https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)"
+          - "This feature is available starting from Prometheus v2.43.0."
```
gardar, line 115:
```suggestion
```

roles/prometheus/meta/argument_specs.yml
```diff
@@ -150,6 +151,14 @@ argument_specs:
         default:
           - "prometheus/targets/*.yml"
           - "prometheus/targets/*.json"
+      prometheus_scrape_config_files:
+        description:
+          - "List of folders where ansible will look for files containing custom scrape config configuration files which will be copied to C({{ prometheus_config_dir }}/scrapes/)."
```
gardar, line 156:
```suggestion
          - "List of folders where ansible will look for files containing custom scrape config configuration files which will be copied to C({{ prometheus_config_dir }}/scrapes/)."
           - "This feature is available starting from Prometheus v2.43.0."
```

gardar (changes requested):
Looks like this line got placed under the wrong option

roles/prometheus/meta/argument_specs.yml
manzsolutions-lpr: Snap, that one’s particularly embarrassing. Sorry and thank you for pointing it out!

gardar (approved):
LGTM! Thanks!

PR #373, opened 2024-05-31
docs: fix config_flags_extra wrong example 

gibix:
The prometheus configuration `storage.tsdb.retention` should be setted via `prometheus_storage_retention`. If setted via
`prometheus_config_flags_extra` it throws `duplicating configuration` error.

### DIFF
--- a/roles/prometheus/meta/argument_specs.yml
+++ b/roles/prometheus/meta/argument_specs.yml
@@ -68,7 +68,7 @@ argument_specs:
       prometheus_config_flags_extra:
         description:
           - "Additional configuration flags passed to prometheus binary at startup"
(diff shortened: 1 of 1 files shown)


gardar:
Nice catch, we are actually making sure that this is not incorrectly configured in the preflight so it makes no sense to have it incorrect in the docs 😅 
https://github.com/prometheus-community/ansible/blob/0800d53d06c9a0fd761c741a250a890b398e14e6/roles/prometheus/tasks/preflight.yml#L54-L67

gardar (approved):
Nice catch, we are actually making sure that this is not incorrectly configured in the preflight so it makes no sense to have it incorrect in the docs 😅 
https://github.com/prometheus-community/ansible/blob/0800d53d06c9a0fd761c741a250a890b398e14e6/roles/prometheus/tasks/preflight.yml#L54-L67

PR #395, opened 2024-07-01
fix: remove EOL platforms for current ones

gardar:


SuperQ (approved):

SuperQ:
It seems like older versions of Ansible fail on some of the newer platforms.

gardar:
> It seems like older versions of Ansible fail on some of the newer platforms.

Yep, I was thinking about adding exclusions to the matrix to skip them. With that we could also trim to list even further by being more explicit about what Ansible versions to test against which distro versions. Just hadn't gotten around to implement it yet.

SuperQ:
Looks like all tests are passing now.

gardar:
Yep! Had to do a few changes as you can see haha.
But basically, now we can define excluded Ansible versions in `.config/molecule/config.yml` 
Also I was able to merge the `prepare` stages in the alternative molecule scenarios for all the roles into one, and do some other minor improvements.

Issue #403, opened 2024-07-17
cadvisor add --env_metadata_whitelist option


BryanQuigley:
From [cadvisor runtime options](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md).

I'm just looking for a way to specify **--env_metadata_whitelist** , which I believe would follow a similar pattern to [enable_metrics](https://github.com/search?q=repo%3Aprometheus-community%2Fansible%20cadvisor_enable_metrics&type=code)

Aside: Upon reviewing the [service file](https://github.com/prometheus-community/ansible/blob/main/roles/cadvisor/templates/cadvisor.service.j2), I do note that these other options are also not implemented - but I get  waiting for someone to need them specifically.
- --store_container_labels=false
- --whitelisted_container_labels
- --raw_cgroup_prefix_whitelist
- --disable_root_cgroup_stats=false
- --container_hints=
- --enable_load_reader=false

PR #407, opened 2024-08-02
enhancement: adding env/container labels options

BryanQuigley:
Add following
- cadvisor_env_metadata_whitelist
- cadvisor_whitelisted_container_labels
- cadvisor_store_container_labels

Tested with this simple playbook:
```
---
- name: Install cAdvisor
  hosts: localhost
  become: yes
  tasks:
    - name: Import cAdvisor role
      import_role:
        name: cadvisor
      vars:
        cadvisor_docker_only: true
        cadvisor_enable_metrics: [ "cpu", "memory", "network", "oom_event" ]
        cadvisor_whitelisted_container_labels: [ "com.docker.compose.oneoff" ]
        cadvisor_env_metadata_whitelist: [ "JEKYLL_ENV", "PATH" ]
        cadvisor_store_container_labels: false
```

Also tested dropping the options, which is why I decided to explicitly set store_container_labels: true (which is the cadvisor default).

Closes: #403

### DIFF
--- a/roles/cadvisor/meta/argument_specs.yml
+++ b/roles/cadvisor/meta/argument_specs.yml
@@ -64,6 +64,18 @@ argument_specs:
         default: []
         type: "list"
         choices: *metrics_choices
+      cadvisor_store_container_labels:
+        description: "store all container labels"
+        type: "bool"
+        default: true
+      cadvisor_whitelisted_container_labels:
+        description: "comma-separated list of container labels to be used as labels on prometheus metrics"
+        default: []
+        type: "list"
+      cadvisor_env_metadata_whitelist:
+        description: "comma-separated list of env variables to be used as labels on prometheus metrics"
+        default: []
+        type: "list"
       cadvisor_docker_only:
         description: "do not report raw cgroup metrics, except the root cgroup"
         type: "bool"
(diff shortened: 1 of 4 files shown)


BryanQuigley:
The failures seem unrelated to the PR - failing to pull image.

BryanQuigley:
@gardar are you the right person to review? Thanks!

roles/cadvisor/meta/argument_specs.yml
```diff
@@ -64,6 +64,18 @@ argument_specs:
         default: []
         type: "list"
         choices: *metrics_choices
+      cadvisor_store_container_labels:
+        description: "do not store all container labels"
+        type: "bool"
+        default: false
```
gardar, line 70: Shouldn't this be
```suggestion
        default: true
```
as defined in `defaults/main.yml`? Or should `defaults/main.yml` perhaps be set to `false`?

gardar (changes requested):

roles/cadvisor/meta/argument_specs.yml
BryanQuigley: Sorry, you are correct. Updating description too

gardar (approved):
LGTM!

BryanQuigley:
Thanks! Just wanted to double check if any other action is needed before merging.

gardar:
> Thanks! Just wanted to double check if any other action is needed before merging.

No sorry, I just needed to fix up the CI config a bit for the tests to pass successfully.

PR #408, opened 2024-08-06
feat: include influxdb exporter role

dobbi84:
Include the [Influxdb exporter](https://github.com/prometheus/influxdb_exporter) role.

We rely on such exporter because we have applications that push metrics directly to Influxdb at runtime, using this exporter allows us to have those metrics in Prometheus. We have created this role getting inspired by the current Node Exporter role already present in this collection, thanks.


SuperQ (changes requested):
This also needs the testing boilerplate under `/tests/integration/targets/`.

SuperQ:
Can you please rebase this against main to pick up the changes in #395

roles/influxdb_exporter/molecule/alternative/prepare.yml
gardar, line 1: Please use the shared prepare step instead, see comment on the `alternative/molecule.yml` file.

roles/influxdb_exporter/molecule/alternative/molecule.yml
```diff
@@ -0,0 +1,10 @@
+---
+provisioner:
```
gardar, line 2: Please include the shared prepare step here

```suggestion
provisioner:
  playbooks:
    prepare: "${MOLECULE_PROJECT_DIRECTORY}/../../.config/molecule/alternative/prepare.yml"
```

roles/influxdb_exporter/vars/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+go_arch_map:
+  i386: '386'
+  x86_64: 'amd64'
+  aarch64: 'arm64'
+  armv7l: 'armv7'
+  armv6l: 'armv6'
+
+go_arch: "{{ go_arch_map[ansible_architecture] | default(ansible_architecture) }}"
```
gardar, line 9: Please use the following instead, so that the variable name is prefixed with the role name.
```suggestion
_influxdb_exporter_go_ansible_arch: "{{ {'i386': '386',
                                'x86_64': 'amd64',
                                'aarch64': 'arm64',
                                'armv7l': 'armv7',
                                'armv6l': 'armv6'}.get(ansible_architecture, ansible_architecture) }}"
```

gardar (changes requested):
Please make sure you follow the following tag pattern for individual tasks:

```
- <role name>
- <purpose>
- <role name>_<purpose>
```

For example:

```
- influxdb_exporter
- download
- influxdb_exporter_download
```

This allows the flexibility of choosing to skip all download tasks in any roles in your playbook with `--skip-tags download`, or just skip the download tasks for this specific role `--skip-tags influxdb_exporter_download`.

roles/influxdb_exporter/meta/argument_specs.yml
```diff
@@ -0,0 +1,67 @@
...
+    author:
+      - "Prometheus Community"
+    options:
+      influxdb_exporter_version:
+        description: "influxdb exporter package version. Also accepts latest as parameter."
+        default: "0.11.5"
+      influxdb_exporter_skip_install:
+        description: "influxdb exporter installation tasks gets skipped when set to true."
+        type: bool
+        default: false
```
gardar, line 17:
```suggestion
```

We don't support changing the install behavior with a variable anymore, users can use tags instead, such as `--skip tags install`

roles/influxdb_exporter/meta/argument_specs.yml
```diff
@@ -0,0 +1,67 @@
...
+          - "System group for influxdb exporter"
+        default: "influxdb-exp"
+      influxdb_exporter_system_user:
+        description:
+          - "I(Advanced)"
+          - "influxdb exporter user"
+        default: "influxdb-exp"
+      influxdb_exporter_archive_path:
+        description: 'Local path to stash the archive and its extraction'
+        default: "/tmp"
```
gardar, line 67:
```suggestion
      influxdb_exporter_local_cache_path:
        description: 'Local path to stash the archive and its extraction'
        default: "/tmp/influxdb_exporter-{{ ansible_system | lower }}-{{ _influxdb_exporter_go_ansible_arch }}/{{ influxdb_exporter_version }}""
```

roles/influxdb_exporter/meta/main.yml
```diff
@@ -0,0 +1,29 @@
...
+      versions:
+        - "bullseye"
+    - name: "EL"
+      versions:
+        - "8"
+        - "9"
+    - name: "Fedora"
+      versions:
+        - "39"
+        - "40"
```
gardar, line 23:
```suggestion
```
#432

roles/influxdb_exporter/tasks/preflight.yml
```diff
@@ -0,0 +1,83 @@
...
+- name: Discover latest version
+  ansible.builtin.set_fact:
+    influxdb_exporter_version: "{{ (lookup('url', 'https://api.github.com/repos/{{ _influxdb_exporter_repo }}/releases/latest', headers=_influxdb_exporter_github_api_headers,
+                            split_lines=False) | from_json).get('tag_name') | replace('v', '') }}"
+  run_once: true
+  until: influxdb_exporter_version is version('0.0.0', '>=')
+  retries: 10
+  when:
+    - influxdb_exporter_version == "latest"
+    - not influxdb_exporter_skip_install
```
gardar, line 53:
```suggestion
```

roles/influxdb_exporter/tasks/preflight.yml
```diff
@@ -0,0 +1,83 @@
...
+        __influxdb_exporter_checksum: "{{ item.split(' ')[0] }}"
+      with_items: "{{ __influxdb_exporter_checksums }}"
+      when:
+        - "('linux-' + go_arch + '.tar.gz') in item"
+  tags:
+    - influxdb_exporter
+    - install
+    - influxdb_exporter_install
+    - download
+    - influxdb_exporter_download
```
gardar, line 83:
```suggestion
```

This is taken care of in the install tasks of the `_common` role

roles/influxdb_exporter/tasks/preflight.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Check if influxdb_exporter is installed
+  ansible.builtin.stat:
+    path: "{{ influxdb_exporter_binary_install_dir }}/influxdb_exporter"
+  register: __influxdb_exporter_is_installed
+  check_mode: false
+  tags:
+    - influxdb_exporter
+    - install
+    - influxdb_exporter_install
```
gardar, line 42:
```suggestion
```

There's no task that's using this information.

gardar (changes requested):

dobbi84:
@gardar it is not allowed to create homes in `/` https://github.com/prometheus-community/ansible/blob/4670e4ad23a081c75fe101ca68fc1f132d9b390e/roles/_common/tasks/install.yml#L40. If a role goes for the default, it fails.

I do not understand why would be needed to set the home to the config dir if anyway `create_home` is set to `false`:
https://github.com/prometheus-community/ansible/blob/4670e4ad23a081c75fe101ca68fc1f132d9b390e/roles/_common/tasks/install.yml#L41

gardar:
> @gardar it is not allowed to create homes in `/`
> 
> https://github.com/prometheus-community/ansible/blob/4670e4ad23a081c75fe101ca68fc1f132d9b390e/roles/_common/tasks/install.yml#L40
> . If a role goes for the default, it fails.
> 
> I do not understand why would be needed to set the home to the config dir if anyway `create_home` is set to `false`:
> 
> https://github.com/prometheus-community/ansible/blob/4670e4ad23a081c75fe101ca68fc1f132d9b390e/roles/_common/tasks/install.yml#L41

Yeah that doesn't sound right, shouldn't it be using the config dir? "{{ influxdb_exporter_config_dir }}"

gardar (approved):
Looks like all the tests are passing now 🥳! Can you please fix the DCO so we can get this merged?

dobbi84:
@SuperQ something is wrong with Molecule Ubuntu container

gardar:
The CI issues should be fixed now #440

PR #411, opened 2024-08-19
minor: Update supported Ansible versions

SuperQ:
Ansible versions 2.11 and older are not supportable on new platforms. Set the minimum supported Ansible version to 2.12 in order to make updating our integration testing possible.
* Fixup minor linting issues.

### DIFF
--- a/.config/molecule/config.yml
+++ b/.config/molecule/config.yml
@@ -17,12 +17,6 @@ platforms:
     privileged: true
     cgroup_parent: docker.slice
     command: /lib/systemd/systemd
-  - name: centos-7
-    image: dokken/centos-7
-    pre_build_image: true
-    privileged: true
-    cgroup_parent: docker.slice
-    command: /usr/lib/systemd/systemd
   - name: centos-stream-9
     image: dokken/centos-stream-9
     pre_build_image: true
--- a/.github/scripts/collection_version_parser.py
+++ b/.github/scripts/collection_version_parser.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 """
 This module provides functions for parsing PEP 440 compliant
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,2 +1,2 @@
 ---
-requires_ansible: ">=2.9.0,<=2.16.99"
+requires_ansible: ">=2.12.0,<=2.16.99"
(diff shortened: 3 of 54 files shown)


SuperQ:
@gardar Fixed up, removing centos-7 here instead of #395 to make some tests pass.

gardar (commented):
I was hoping centos 7 would be the last version using 2.9 but I don't think we can fully remove the old ansible versions yet, as for example ubuntu 20.04 doesn't have anything newer than 2.9 in the repos, not even in the backports https://packages.ubuntu.com/focal/ansible 😔 
I'd much prefer that we add exclusions to the matrix as I mentioned here: https://github.com/prometheus-community/ansible/pull/395#issuecomment-2297472476

SuperQ:
Ok, if you have a hint on how/where you were going to add the version matrix list, I could work on it.

gardar:
I'm not absolutely certain where it would be best to place the exclusion list, I don't think that it would fit in any of the ansible meta files. So I guess it would either have to be under `env:` in `.github/workflows/ansible-ci.yml` on in some new file under `.github/`

Something like this for example:
```yaml
env:
  EXCLUSIONS: |
            - target: ubuntu-24.04
              ansible-core-version: 
                - stable-2.10
                - stable-2.11
```

Then we need to add it to the matrix json that's dynamically generated and used in `.github/workflows/ansible-test-molecule.yml`

The matrix that's currently generated ends something like this:

```yaml
jobs:
  ansible-test-integration:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      max-parallel: 4
      matrix:
        target: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
        ansible-core-version: [stable-2.10, stable-2.11, stable-2.12, stable-2.13]
```

but after adding the excludes it would end something like this:

```yaml
jobs:
  ansible-test-integration:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      max-parallel: 4
      matrix:
        target: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
        ansible-core-version: [stable-2.10, stable-2.11, stable-2.12, stable-2.13]
        exclude:
          - target: ubuntu-24.04
            ansible-core-version: 
              - stable-2.10
              - stable-2.11
```

SuperQ:
Closing in favor of https://github.com/prometheus-community/ansible/pull/416.

PR #421, opened 2024-09-30
fix(process_exporter): skip install when there's nothing to do

frazew:
Just like in `roles/node_exporter/tasks/main.yml`: the install tasks can be skipped conditionally when process_exporter is already installed and reports the same version as the expected one.

This short patch is basically a copy-paste from [roles/node_exporter/tasks/main.yml](https://github.com/prometheus-community/ansible/blob/main/roles/node_exporter/tasks/main.yml#L22) with the variables updated.


gardar:
Obsoleted by #425

PR #424, opened 2024-10-15
removed: variables for skipping install in favor of tags

gardar:
Using a variable to skip installation is redundant because the same result can be achieved with tags. Tags already provide the flexibility to skip the installation or other sections of the roles, such as configuration tasks.


gardar:
> Should we add something to the docs about this?

Not sure, it's fairly standard practice to use tags in ansible. I added a assert task to the preflight which instructs users to use `--skip-tags` instead of the variable if they have the variable defined.

gardar:
superseded by #425

PR #425, opened 2024-10-15
refactor: consolidate common tasks

### LINKED ISSUE #64: Share common tasks across roles

gardar:
This PR refactors the roles to take advantage of being in a collection by consolidating shared tasks into a new internal `_common` role. Previously many of the roles were maintained as standalone with some unavoidable redundancy.

By centralizing shared tasks, we improve consistency and reduce the maintenance burden across the collection.

Note that it's not the plan to have the roles as meta roles by moving everything over to the `_common` role but to use the `_common` role for tasks that are present in all/most of the roles.

fixes #64

### DIFF
--- a/.github/workflows/ansible-test-integration.yml
+++ b/.github/workflows/ansible-test-integration.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      max-parallel: 4
+      max-parallel: 10
       matrix:
         targets: ${{ fromJson(inputs.targets) }}
         ansible-core-versions: ${{ fromJson(inputs.ansible-core-versions) }}


SuperQ (approved):
Amazing!

PR #432, opened 2024-10-18
removed: CentOS Stream and Fedora from being officially supported

gardar:
Removed CentOS Stream and Fedora from being officially supported.

- CentOS Stream should still work as we are running tests on AlmaLinux.
- Fedora isn't widely used as a server distro

This should trim down our CI test count/time a bit.

### DIFF
--- a/.config/molecule/config.yml
+++ b/.config/molecule/config.yml
@@ -17,12 +17,6 @@ platforms:
     privileged: true
     cgroup_parent: docker.slice
     command: /lib/systemd/systemd
-  - name: centos-stream-9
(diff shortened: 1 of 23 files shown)


gardar:
> I kinda feel like it would be nice to keep Fedora, since it's a nice canary for new upstream releases code if we keep it up-to-date with the latest Fedora versions.
> 
> But this is fine for now.

Good point! 
Perhaps we could take it even further and test upcoming/pre releases of Prometheus and exporters with fedora/archlinux/whatever in a new molecule scenario?

PR #438, opened 2024-10-23
fix(blackbox_exporter): make capability setting idempotent

sverrehu:
Recently, the "Ensure blackbox exporter binary has cap_net_raw capability" task has reported a change for every pipeline run. This PR contains two changes to fix this behavior.

* An [old bug](https://github.com/ansible-collections/community.general/issues/4067) in `community.general.capabilities`, which is caused by naive comparison between requested state and observed state, will indicate a change if the operation is anything besides `=`. This PR changes from `cap_net_raw+ep` to `cap_net_raw=ep` to fix this.

* A [recent PR](https://github.com/prometheus-community/ansible/blob/main/roles/blackbox_exporter/tasks/configure.yml#L22) set `changed_when: "'molecule-idempotence-notest' not in ansible_skip_tags"`, which translates to `changed_when: true`, causing this task to appear as changed no matter what. This PR removes this line.

With these changes, our pipeline reports no change if the `cap_net_raw` is as wanted, and reports change otherwise.

### DIFF
--- a/roles/blackbox_exporter/tasks/configure.yml
+++ b/roles/blackbox_exporter/tasks/configure.yml
@@ -15,11 +15,10 @@
 - name: Ensure blackbox exporter binary has cap_net_raw capability
   community.general.capabilities:
     path: "{{ blackbox_exporter_binary_install_dir }}/blackbox_exporter"
-    capability: cap_net_raw+ep
+    capability: cap_net_raw=ep
     state: present
   become: true
   when: not ansible_check_mode
-  changed_when: "'molecule-idempotence-notest' not in ansible_skip_tags"
   tags:
     - blackbox_exporter
     - configure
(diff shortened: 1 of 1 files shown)


SuperQ:
This wouldn't case Prometheus to restart, only the blackbox_exporter.

SuperQ:
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

gardar (approved):

sverrehu:
I noticed some idempotence checks failed. Previously, those tests would en effect have `changed_when: false`, as `molecule-idempotence-notest` was part of `ansible_skip_tags` in Molecule's idempotence checks.

I just added `molecule-idempotence-notest` as part of the `tags`, in order to skip this task completely. Hopefully, the effect should be the same as before.

gardar:
The failing CI tests will be fixed with #440

gardar (approved):

PR #443, opened 2024-10-29
refactor: move naive assertion to common role

zhan9san:
If this proposal is accepted, I'll update all other roles

### DIFF
--- a/roles/smartctl_exporter/tasks/preflight.yml
+++ b/roles/smartctl_exporter/tasks/preflight.yml
@@ -3,22 +3,14 @@
   ansible.builtin.include_role:
     name: prometheus.prometheus._common
     tasks_from: preflight.yml
+  vars:
+    _common_web_listen_address: "{{ smartctl_exporter_web_listen_address }}"
 
 - name: Assert that used version supports listen address type
   ansible.builtin.assert:
     that:
       - >-
         smartctl_exporter_web_listen_address is string
-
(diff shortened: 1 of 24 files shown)


SuperQ (approved):

gardar (changes requested):
Thanks! Please go ahead with applying the change across all the roles

zhan9san:
Done. It's ready for review

roles/smartctl_exporter/tasks/preflight.yml
gardar, line 1: Don't forget to remove the naive assertion task here

gardar (changes requested):

roles/smartctl_exporter/tasks/preflight.yml
zhan9san: Good point. It's removed.

zhan9san:
> I wonder if this assert should be added to the cadvisor role too.

I should list cadvisor in the description.

It is not added intentionally. Because it is not in standard format.

```shell
$ docker run --rm -it gcr.io/cadvisor/cadvisor --help | grep -E 'listen|port'
    	Only report docker containers in addition to root stats
  -listen_ip string
    	IP to listen on, defaults to all IPs
  -port int
    	port to listen (default 8080)
```

`node_exporter` and others is in this format, `--web.listen-address=:9100`.

gardar (approved):
LGTM

Issue #147, opened 2023-06-19
Custom alerts are not being removed


sergiocharpineljr:
Currently there is no easy way to remove custom rules. Since we are using [copy](https://github.com/prometheus-community/ansible/blob/main/roles/prometheus/tasks/configure.yml#L16), if we have any local rule, it won't be removed in the server.

zhan9san:
@sergiocharpineljr 

Good point. It is not idempotent currently.

Would you please help implement it like how Grafana does?

Reference:

https://github.com/grafana/grafana-ansible-collection/blob/84d6af3b1c3991d6c560a8fe9c09d8fc1811f520/roles/grafana/tasks/dashboards.yml#L136-L142

PR #453, opened 2024-11-06
feat: Add prometheus prometheus_provisioning_synced support

sergiocharpineljr:
Add support for prometheus_provisioning_synced (fixes https://github.com/prometheus-community/ansible/issues/147).


sergiocharpineljr:
I tested this locally but didn't include any tests. If you feel like we should, please let me know and give me some guidance on the best approach to follow.

gardar (commented):
Thanks. 
I wonder if it might be cleaner to simply switch the `Copy custom alerting rule files` task over to the `ansible.posix.synchronize` module with `delete: true` to solve this.
Also generally not a fan of using the `json_query` filter unless absolutely necessary.

sergiocharpineljr:
> Thanks. I wonder if it might be cleaner to simply switch the `Copy custom alerting rule files` task over to the `ansible.posix.synchronize` module with `delete: true` to solve this. Also generally not a fan of using the `json_query` filter unless absolutely necessary.

I've followed the suggestion in https://github.com/prometheus-community/ansible/issues/147 to base this on Grafana's code: https://github.com/grafana/grafana-ansible-collection/blob/84d6af3b1c3991d6c560a8fe9c09d8fc1811f520/roles/grafana/tasks/dashboards.yml#L136-L142

The `ansible.posix.synchronize` might be tricky because of the `ansible_managed.yml` file we have in the same dir. So not sure how to make this idempotent.

`json_query` was used because of the `default` filter. Doing `__rules_managed_copied.dest` directly causes issues when `prometheus_alert_rules = []` and the `__rules_managed_copied` var is not created.

zhan9san:
hi @sergiocharpineljr 

`ansible.posix.synchronize` can work as well. 

FYI

```yaml
---
- name: Synchronize directories using ansible.posix.synchronize
  hosts: localhost
  tasks:
    - name: Synchronize foo/rules to bar/rules with delete option
      ansible.posix.synchronize:
        src: ~/tmp/foo/rules/
        dest: ~/tmp/bar/rules/
        delete: yes
        recursive: yes
        rsync_opts:
          - "--exclude='ansible_managed.yml'"
```

```shell
jack@v:~/tmp$ ansible-playbook a.yaml
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [Synchronize directories using ansible.posix.synchronize] **********************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] **************************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Synchronize foo/rules to bar/rules with delete option] ************************************************************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP **************************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

jack@v:~/tmp$ tree foo/rules/ bar/rules/
foo/rules/
└── aaa.yml
bar/rules/
├── aaa.yml
└── ansible_managed.yml

2 directories, 3 files
```

sergiocharpineljr:
@zhan9san @gardar Ok, I can try the `synchronize` approach. Should we leave it as optional, keeping the current method as default (or setting `delete: false`) or do you think we should always force `copy` instead?

zhan9san:
I prefer using `synchronize ` with `delete: true` as default.

@gardar  Your thought?

gardar:
> I prefer using `synchronize ` with `delete: true` as default.
> 
> @gardar Your thought?

I agree; ideally, the roles should strictly manage how Prometheus is configured. However, I understand that’s not always the case, so having it as an optional feature could be useful.

Another concern I have is that using the synchronize module adds rsync as a dependency to the role, and rsync isn’t always installed by default.

sergiocharpineljr:
> > I prefer using `synchronize ` with `delete: true` as default.
> > @gardar Your thought?
> 
> I agree; ideally, the roles should strictly manage how Prometheus is configured. However, I understand that’s not always the case, so having it as an optional feature could be useful.
> 
> Another concern I have is that using the synchronize module adds rsync as a dependency to the role, and rsync isn’t always installed by default.

Yes, well remembered. So @zhan9san @gardar can we stick with the original proposal?

PR #461, opened 2024-11-17
skip_changelog: Fix role repo discovery

SuperQ:
Fix syntax of split to correctly select list item.
* Add some extra stderr debugging logs.
* Skip the `_common` role.
* Skip roles where the repo is missing.


SuperQ:
This will fix the updater cron job.

gardar (approved):
Nice!

Issue #460, opened 2024-11-15
"only prometheus 2.x is supported" for `prometheus_version`


dmke:
Does this note only exists to discourage user to install Prometheus 1.x with the `prometheus.prometheus.prometheus` role, or is it actually not possible to install Prometheus 3.0 with it?

> ![image](https://github.com/user-attachments/assets/e44dccf4-034f-414c-89a8-5efe146372b9)
> 
> https://prometheus-community.github.io/ansible/branch/main/prometheus_role.html

I'm hesitant to try this on my production machine, in case something unexpected breaks, hence this request for clarification.

SuperQ:
Yes, this was originally to discourage Prometheus 1.x.

I don't think there are any issues with 3.x, but we haven't explicitly tested it yet.

dmke:
OK, thanks. I'll later setup a test machine, try to deploy 3.x to it, and report back.

PR #462, opened 2024-11-17
feat: Support Prometheus 3.0

SuperQ:
Update prometheus role for 3.0.0.
* Update argument_specs.yml to reflect support of 3.x.
* Skip console templates on 3.x.

Fixes: https://github.com/prometheus-community/ansible/issues/460

### DIFF
--- a/roles/prometheus/tasks/configure.yml
+++ b/roles/prometheus/tasks/configure.yml
@@ -36,6 +36,8 @@
     group: "{{ prometheus_system_group }}"
     mode: 0775
   loop:
+    - "{{ prometheus_config_dir }}/consoles"
+    - "{{ prometheus_config_dir }}/console_libraries"
     - "{{ prometheus_config_dir }}/rules"
     - "{{ prometheus_config_dir }}/file_sd"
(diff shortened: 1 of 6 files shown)


SuperQ:
I went over https://prometheus.io/docs/prometheus/3.0/migration/ and didn't find anything incompatible in our role.

roles/prometheus/tasks/configure.yml
```diff
@@ -58,6 +60,7 @@
   notify:
     - restart prometheus
   become: true
+  when: prometheus_version is version_compare('3.0.0', '<')
```
gardar, line 63: Should probably be just version and not version_compare, afaik the version_compare name has been deprecated

SuperQ: Ahh, that was copy-pasta from other roles. Will update it.

SuperQ:
Found and fixed a couple of minor issues.
* It doesn't come with console templates anymore, but still supports them.
* It renamed the agent feature flag to just a flag.

gardar (approved):
LGTM!

PR #474, opened 2024-11-21
fix: use ansible_facts instead of ansible_* variables

### LINKED ISSUE #297: Allow roles to run with INJECT_FACTS_AS_VARS set to false
Currently, the roles failed to run with `INJECT_FACTS_AS_VARS` set to `false` as the required `ansible_*` variables (e.g. `ansible_os_family`) are not defined.

The configuration `INJECT_FACTS_AS_VARS` and the Ansible fact namespace `ansible_facts.*` has been added in Ansible 2.5. In the [porting guide of that version](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.5.html#ansible-fact-namespacing) it stated that:

> A new configuration variable, `inject_facts_as_vars`, has been added to `ansible.cfg`. Its default setting, ‘True’, keeps the 2.4 behavior of facts variables being set in the old `ansible_*` locations (while also writing them to the new namespace). This variable is expected to be set to ‘False’ in a future release. When inject_facts_as_vars is set to False, you must refer to `ansible_facts` through the new `ansible_facts.*` namespace.

[It was also confirmed by Ansible developer on reddit](https://www.reddit.com/r/ansible/comments/13giahm/comment/jkakft9/?utm_source=share&utm_medium=web2x&context=3) that `INJECT_FACTS_AS_VARS` was planned to be deprecated in the future.

Therefore, it would be great if the collections supported `INJECT_FACTS_AS_VARS=false` as well.

To fix this just change all `ansible_*` variables to `ansible_facts.*`. As the minimum supported ansible version of this collection is 2.9 I expect it should continue to work after the change. I can do a PR when I have time.

kleini:
to allow ANSIBLE_INJECT_FACT_VARS to become false.

fixes #297

### DIFF
--- a/.config/molecule/config.yml
+++ b/.config/molecule/config.yml
@@ -48,6 +48,8 @@ verifier:
   env:
     PYTHONPATH: "${MOLECULE_PROJECT_DIRECTORY}/../../.testinfra:${PYTHONPATH}"
 provisioner:
+  env:
+    ANSIBLE_INJECT_FACT_VARS: false
   playbooks:
     converge: "${MOLECULE_PROJECT_DIRECTORY}/../../.config/molecule/converge.yml"
   inventory:
--- a/roles/_common/meta/argument_specs.yml
+++ b/roles/_common/meta/argument_specs.yml
@@ -77,8 +77,8 @@ argument_specs:
     options:
       _common_dependencies:
         description: "Package dependencies to install"
-        default: "{% if (ansible_pkg_mgr == 'apt') %}\
-                  {{ ('python-apt' if ansible_python_version is version('3', '<') else 'python3-apt') }}
+        default: "{% if (ansible_facts['pkg_mgr'] == 'apt') %}\
+                  {{ ('python-apt' if ansible_facts['python_version'] is version('3', '<') else 'python3-apt') }}
                   {% else %}\
                   {% endif %}"
       _common_web_listen_address:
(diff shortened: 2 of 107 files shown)


kleini:
I have no idea about the failing fail2ban roles. This seems to be broken in every pull request. Can somebody give me a hint, how to fix that? Or can this be fixed on the `main` branch?

mgariepy:
> I have no idea about the failing fail2ban roles. This seems to be broken in every pull request. Can somebody give me a hint, how to fix that? Or can this be fixed on the `main` branch?

seems like 308 error are not handled in fail2ban check prior to version 3.11.0 python version 

https://github.com/python/cpython/commit/c379bc5ec9012cf66424ef3d80612cf13ec51006

kleini:
If that is only fixed in Python 3.11, tests for Ansible <2.14 can not be fixed as they do not support that new Python version.

mgariepy:
> If that is only fixed in Python 3.11, tests for Ansible <2.14 can not be fixed as they do not support that new Python version.

the 308 redirect is kinda new, since it seems only to be done for fail2ban 0.10.2 released on November 20th.

downloading the release 0.10.1 checksums only goes through a 301.

mgariepy:
here is the link to the gitlab issue : https://gitlab.com/gitlab-org/gitlab/-/issues/508161

kleini:
Rebased to include #507. Now everything's green.

gardar (approved):
LGTM! Thanks

Issue #478, opened 2024-11-27
Missing consoles and console_libraries files when installing Prometheus 3.0.0


maximepoulin-hl:
Since upgrading to version 0.23.0, `prometheus.prometheus.prometheus` fails to install Prometheus version 3.0.0 with the following error:

```
TASK [prometheus.prometheus.prometheus : Propagate official console templates] *****************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: If you are using a module and expect the file to exist on the remote, see the remote_src option
failed: [prometheus1] (item=console_libraries) => {"ansible_loop_var": "item", "changed": false, "item": "console_libraries", "msg": "Could not find or access '/tmp/prometheus-linux-arm64/3.0.0/console_libraries' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: If you are using a module and expect the file to exist on the remote, see the remote_src option
failed: [prometheus1] (item=consoles) => {"ansible_loop_var": "item", "changed": false, "item": "consoles", "msg": "Could not find or access '/tmp/prometheus-linux-arm64/3.0.0/consoles' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
```

I've checked and the downloaded tarball is indeed missing those files. The directories are there on 2.55.1.

Works as expected installing Prometheus 2.55.1.

SuperQ:
This is an expected change. Prometheus no longer includes consoles in 3.0.

SuperQ:
Wait, sorry, that should not error. It should skip the whole step.

PR #481, opened 2024-11-27
fix: Use lookup for console template test

SuperQ:
When deploying the console templates, use a local lookup to check if the Prometheus tarball contains consoles to deploy.


maximepoulin-hl:
I'm surprised `when: prometheus_version is version('3.0.0', '<')` didn't work, I had that explicitly set to 3.0.0 in #478 so I'm not sure why I still ended up running that task.

Issue #484, opened 2024-12-03
Prometheus custom rules not copying to rules folder


cdestouni-fmp:
## Setup

I upgraded to the latest release v0.23.0 and run the Prometheus role to configure Prometheus on a Ubuntu 22 virtual machine. My custom rules are located in `prometheus/rules/*yml` location and my files all end with the `.yml` extension as expected by the default variable `prometheus_alert_rules_files`. 

### Configuration

I 'm running the role using the below ansible configuration.

```
ansible [core 2.13.13]
  config file = /home/user/fmp_git/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/.local/bin/ansible
  python version = 3.8.10 (default, Nov  7 2024, 13:10:47) [GCC 9.4.0]
  jinja version = 3.0.3
  libyaml = True
```

And the below OS configuration

```
Distributor ID:	Ubuntu
Description: Ubuntu 20.04.5 LTS
Release: 20.04
Codename: focal
```

Deploying to the below OS configuration

```
Distributor ID:	Ubuntu
Description: Ubuntu 22.04.3 LTS
Release: 22.04
Codename: jammy
```

## Problem 

The [Copy custom alerting rule files](https://github.com/prometheus-community/ansible/blob/0.23.0/roles/prometheus/tasks/configure.yml#L88) did not copy over my custom rules to the rules directory on the Prometheus server. There was no feedback, even in verbose mode `-vvvv` so I assumed the glob working correctly and it thinks there's no files to copy over. After the role completed, the Prometheus server was running but no custom rules were displayed in the UI.

## Testing done

- I tried to override the `prometheus_alert_rules_files` variable with a new location (In the expected list format). This resulted in an error of not locating the files instead of the original no feedback, which makes me think the location is not the problem.

- I tested at v0.19.0 before this task was last changed and it works fine and copied the custom rules over as expected.

- I tried the v0.23.0 task changes but using `root` permissions on the copy task which did not make a difference. This means it is not a permissions issue.

## Conclusion 

This leaves me to think the problem is situated in the change below.

Working version (v0.19.0)

```yml
with_fileglob: "{{ prometheus_alert_rules_files }}"
```
Not working version (v0.23.0)

[...]

PR #488, opened 2024-12-05
fix: use with_fileglob for iterating over custom alerting rule files

MajorP93:
Fixes https://github.com/prometheus-community/ansible/issues/484

### DIFF
--- a/roles/prometheus/tasks/configure.yml
+++ b/roles/prometheus/tasks/configure.yml
@@ -93,7 +93,7 @@
     group: "{{ prometheus_system_group }}"
     mode: 0640
     validate: "{{ prometheus_binary_install_dir }}/promtool check rules %s"
-  loop: "{{ prometheus_alert_rules_files | map('ansible.builtin.fileglob') | flatten }}"
+  with_fileglob: "{{ prometheus_alert_rules_files }}"
   when:
     - not prometheus_agent_mode
   notify:


gardar:
> I fixed the issue using an absolute path when configuring `prometheus_alert_rules_files`, like this:
> 
> 
> 
> ```
> 
> prometheus_alert_rules_files:
> 
>   - "{{ playbook_dir }}/roles/prometheus/prometheus/rules/generic/*.yml"
> 
> ```
> 
> The other thing I noticed is that doing `{{ prometheus_alert_rules_files | map('ansible.builtin.fileglob') | flatten` returns relative paths if you configure `prometheus_alert_rules_files` with relative paths, which is problematic with the use of ansible.builtin.copy. When using `with_fileglob`, the paths returned are always absolute. 

 _Originally posted by @lsimard in [#484](https://github.com/prometheus-community/ansible/issues/484#issuecomment-2573780353)_

weakcamel:
@gardar Curious, why has this PR been deemed invalid?

The comment in https://github.com/prometheus-community/ansible/issues/484#issuecomment-2573780353 is a good workaround but still a workaround; using `with_fileglob` would IMO solve this issue properly.

MajorP93:
I was wondering the same. I use my own fork in our internal company network which uses `with_fileglob` as it works as expected and reliable.
Never understood why maintainers went with `loop: "{{ prometheus_alert_rules_files | map('ansible.builtin.fileglob') | flatten }}" ` instead as it seems to create more problems than it solves.

PR #498, opened 2024-12-20
fix: use throttle: 1 for localhost unarchive

### LINKED ISSUE #457: Concurrent local extraction of binaries fails sometimes
When the node_exporter role is used to install the exporter on > 1 hosts we see occasional failures of binary extraction after download like so:

```
TASK [prometheus.prometheus._common : Unpack binary archive node_exporter-1.8.2.linux-arm64.tar.gz] ***
changed: [52.27.162.6 -> localhost]
changed: [34.212.107.78 -> localhost]
changed: [35.160.16.87 -> localhost]
changed: [34.211.30.255 -> localhost]
fatal: [35.91.204.184 -> localhost]: FAILED! => {"changed": true, "dest": "/tmp/node_exporter-linux-arm64/1.8.2", "extract_results": {"cmd": ["/usr/bin/tar", "--extract", "-C", "/tmp/node_exporter-linux-arm64/1.8.2", "-z", "--show-transformed-names", "--strip-components=1", "-f", "/root/.ansible/tmp/ansible-tmp-1730343705.1365232-1179-87748719038915/source"], "err": "", "out": "", "rc": 0}, "gid": 0, "group": "root", "handler": "TgzArchive", "mode": "0755", "msg": "Unexpected error when accessing exploded file: [Errno 2] No such file or directory: b'/tmp/node_exporter-linux-arm64/1.8.2/node_exporter'", "owner": "root", "size": 81, "src": "/root/.ansible/tmp/ansible-tmp-1730343705.1365232-1179-87748719038915/source", "state": "directory", "uid": 0}
changed: [54.212.243.52 -> localhost]
changed: [35.91.104.132 -> localhost]
```

It's probably less than 1% of total extraction executions, but with many hosts the overall failure rate is quite high.

We delegate the download & extraction to localhost, so this runs in parallel N times for N hosts but the paths are the same in every case, so it's concurrently extracting to the same destination path which I guess causes this, e.g, [from the manual](https://www.gnu.org/software/tar/manual/html_node/Dealing-with-Old-Files.html):

> When extracting files, if tar discovers that the extracted file already exists, it normally replaces the file by removing it before extracting it, to prevent confusion in the presence of hard or symbolic links.

So there will be brief periods where a written file (in this case, the top level dir) will be deleted and replaced by a concurrent extraction and if some earlier writer tries to list it it will see it missing.

Besides the race, it's also inefficient to extract the same file N times.

I guess `run_once` would fix this if it wasn't for the complication of multi-arch. Maybe `serial` execution + skip if already exists would work with little refactoring of the existing approach?

I think this race existed before the big refactor in 0.20.0, but in the refactor `list_files` was added which may be the thing that triggers the specific failure above.

travisdowns:
During the common install tasks, we download and unpack the binary to install on localhost and eventually upload it to each host (among many other things). Currently the "unpack" step, which extracts the gzipped tar archive, is perform N times if there are N hosts in the inventory, but the target directory (something like
/tmp/node_exporter-linux-arm64/1.8.2) is the same for every host with the same architecture.

This means that all the unarchive tasks are extracting in parallel in an unsynchronized manner to the same directory. It a miracle that this works, but at least usually it like it does, but sometimes tar will complain about a missing file, failing the install. I've confirmed locally that this is due to the race described above.

To fix this, we use `throttle: 1` on the unpack task. This means that the first task (for each architecture) will do the unpack and the other tasks are effectively no-ops, they are reported as "ok" rather than changed in the output.

Fixes #457.

### DIFF
--- a/roles/_common/tasks/install.yml
+++ b/roles/_common/tasks/install.yml
@@ -118,6 +118,7 @@
         list_files: true
         extra_opts: "{{ _common_binary_unarchive_opts | default(omit, true) }}"
       check_mode: false
+      throttle: 1  # see prometheus-community/ansible#457
       when: __common_binary_basename is search('\.zip$|\.tar\.gz$')
 
 - name: "Check existence of binary install dir"


travisdowns:
I tested this locally with a file that is more likely to trigger the race: a tar of 10,000 small files all at the top level of the tar: this triggers the error 100% of the time w/o the fix and passes 1,000 iterations locally with no iteration.

I also verified that deploying prom to a 4 host inventory works, 100 iterations without failure. This inventory has 2 hosts each for 2 different architectures (amd64 and arm64).

gardar (approved):
That's a clean and simple solution - I like it! Thanks!

roles/_common/tasks/install.yml
```diff
@@ -102,6 +102,7 @@
         extra_opts: "{{ _common_binary_unarchive_opts | default(omit, true) }}"
       delegate_to: localhost
       check_mode: false
+      throttle: 1 # see prometheus-community/ansible#457
```
SuperQ, line 105:
```suggestion
      throttle: 1  # see prometheus-community/ansible#457
```

travisdowns: Accepted & squashed.

travisdowns:
The test failures look like this:

```
TASK [prometheus.prometheus._common : Get checksum list for fail2ban_exporter_0.10.2_linux_amd64.tar.gz] ***
fatal: [almalinux-9]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'url'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Received HTTP error for https://gitlab.com/hectorjsmith/fail2ban-prometheus-exporter/-/releases/v0.10.2/downloads/fail2ban_exporter_0.10.2_checksums.txt : HTTP Error 308: Permanent Redirect"}
```

I believe this means the external resource has changed. It looks like the user changed their name and the correct link is:


https://gitlab.com/hctrdev/fail2ban-prometheus-exporter/-/releases/v0.10.2/downloads/fail2ban_exporter_0.10.2_checksums.txt

SuperQ:
I think the 308 error is related to https://github.com/prometheus-community/ansible/pull/507

gardar:
Please rebase now that the 308 error should be resolved

travisdowns:
@gardar rebase on tip of main as of today, let's see how things go.

travisdowns:
The CI failure is spurious:

```
TASK [prometheus.prometheus.postgres_exporter : Discover latest version] *******
fatal: [almalinux-9]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'url'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Received HTTP error for https://api.github.com/repos/prometheus-community/postgres_exporter/releases/latest : HTTP Error 403: rate limit exceeded"}
```

travisdowns:
@gardar is there some way to retry just that one CI command which was rate limited? Rather than all 660.

gardar:
> @gardar is there some way to retry just that one CI command which was rate limited? Rather than all 660.

I can re run individual jobs and I did just re-run that failed job, however I'm not aware of any way for contributors to re-run failed jobs.

PR #499, opened 2024-12-21
feature: allow downloading of package by remote host

### LINKED ISSUE #389: ansible: node-exported download behavior
Node exporter downloads locally and then uploads to all hosts. Imagine you are deploying something from your laptop over a dinky home connection to 100 hosts in the cloud. This is going to take a long time.

Can we add a mode where it downloads the binaries onto the hosts themselves? This could be as simple as using `omit` for the `delegate_to:` values in the tasks that deal with with the download and then skipping the upload ("propagate") step.

travisdowns:
Currently node_exporter install always downloads the package on the localhost and unpacks it there then uploads it to the remote host.

In the case the ansible controller is at the other end of a thin network pipe, or if it may not even have access to the package URL, it may be desirable to download and unpack it on the remote host instead.

This change adds that ability as an optional behavior to _common/install and wires it up for use by node_exporter, using the node_exporter_localhost_download variable. The default for node exporter and any other binaries remains the same as before: delegate to localhost.

Fixes #389.


gardar (changes requested):
Thanks!
I think the `_common_delegate_to_localhost` variable name is a bit too generic, how about `_common_remote_download` and `node_exporter_remote_download` etc. instead?
Also I think we should implement this in all the roles at once rather than just for the node_exporter.

travisdowns:
> Also I think we should implement this in all the roles at once rather than just for the node_exporter.

Is there some efficient way to test this? I tested node_exporter (by hand, locally) but I don't think I'll be able to test all 20+ roles which rely on _common.

travisdowns:
> I think the `_common_delegate_to_localhost` variable name is a bit too generic, how about `_common_remote_download` and `node_exporter_remote_download` etc. instead?

Good call, I've applied your suggestion in push https://github.com/prometheus-community/ansible/commit/19cb70e1ac5472a14a51a88d91512cce3470110e.

zhan9san:
@travisdowns 

Awesome. LGTM

travisdowns:
Thanks for the note. I'm not able to pick this back up at the moment but I may return in the future (moving to draft for now).

PR #504, opened 2025-01-05
skip_changelog: Pin almalinux to old container images

SuperQ:
Test if rolling back almalinux fixes build issues.


gardar (approved):
Nice one!
Perhaps we should pin the other containers as well? We can then use renovate bot to ensure that they are kept up to date

SuperQ:
I don't think we actually want to pin them. I'm hoping this is temporary until we figure out why the current version is broken.

PR #507, opened 2025-01-07
fix: unsupported http response code issues on older ansible/python versions

gardar:
GitLab has recently started using 308 redirects for release assets, which causes compatibility issues with older versions of Ansible and Python. ~~Our options for resolving this are limited, so the best solution for now is to allow the checksum download task to proceed with a warning. Unfortunately, this has the side effect of disabling checksum verification when the warning is triggered, but AFAIK it remains the only workable approach at this time.~~
To mitigate the issue I have switched the tasks from using the `get_url` module and the `url` lookup to use the `uri` module instead as it supports configuring which http status codes should be considered as successful and also enables redirect behavior customization.

See more here: 
https://github.com/prometheus-community/ansible/pull/475#issuecomment-2511739134
https://github.com/prometheus-community/ansible/actions/runs/11955696431/job/33328750307
cc @mgariepy

### DIFF
--- a/roles/_common/tasks/install.yml
+++ b/roles/_common/tasks/install.yml
@@ -78,20 +78,44 @@
       run_once: true
       when: (_common_checksums_url)
 
-    - name: "Download {{ __common_binary_basename }}"
-      ansible.builtin.get_url:
-        url: "{{ _common_binary_url }}"
-        dest: "{{ _common_local_cache_path }}/{{ _common_binary_name | default(__common_binary_basename) }}"
-        headers: "{{ __common_github_api_headers }}"
-        checksum: "{{ __common_binary_checksums[__common_binary_basename] | default(omit) }}"
-        mode: 0644
-      register: __common_download
-      until: __common_download is succeeded
-      retries: 5
-      delay: 2
-      # run_once: true  # <-- this can't be set due to multi-arch support
-      delegate_to: localhost
-      check_mode: false
+    - name: "Download binary"
+      block:
+        - name: "Download {{ __common_binary_basename }}"
+          ansible.builtin.get_url:
+            url: "{{ _common_binary_url }}"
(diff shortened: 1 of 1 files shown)


SuperQ:
What versions of python/ansible are unsupported?

SuperQ (approved):
Read through the comments, I guess this is fine for now.

gardar:
Looks like it's ansible 2.9-2.13
https://github.com/prometheus-community/ansible/actions/runs/11955696431/job/33328750307

btw. I was able to get a 302 response from gitlab by forcing http1.0 but unfortunately there doesn't seem to set the http version in the ansible module, I tried various headers to try and trick it to give a http1.0 response but no avail.

mgariepy:
wow, it seems to works i wonder if we could verify checksum in an extra task during the rescue, just to keep integrity.

gardar:
> wow, it seems to works i wonder if we could verify checksum in an extra task during the rescue, just to keep integrity.

Good point! I added a separate task for checksum verification, that enabled me to replace `get_url` in the download task with `uri` so that we don't actually need to add the fallback download for the old ansible versions.

mgariepy:
> > wow, it seems to works i wonder if we could verify checksum in an extra task during the rescue, just to keep integrity.
> 
> Good point! I added a separate task for checksum verification, that enabled me to replace `get_url` in the download task with `uri` so that we don't actually need to add the fallback download for the old ansible versions.

it's also needed when using older python with new-ish ansible.

Thanks a lot for your patches.

gardar:
> it's also needed when using older python with new-ish ansible.

Right, the issue can probably be attributed to the older python libraries rather than Ansible itself, although it would be nice if the ansible modules had parameters to define what HTTP version to use.

> Thanks a lot for your patches.

No problem, thanks to you as well for initially figuring out the root cause of the issue! :smiley:

SuperQ (approved):

PR #515, opened 2025-01-15
feat(snmp_exporter): fix dry-runs on new machines

pieterlexis-tomtom:
This does not fail a dry run on machines where snmp_exporter is not yet installed.


gardar:
Is this really needed? `ansible.builtin.systemd` has full support for check_mode

pieterlexis-tomtom:
> Is this really needed? `ansible.builtin.systemd` has full support for check_mode

From what I have noticed: if you run this role in check mode on a blank slate, the unitfile is not put unto the system. When the service is started (or restarted) the dry-run will error out with `Could not find the requested service snmp_exporter: host`, stopping the dry-run. I can try to do some more experiments.

PR #520, opened 2025-01-20
fix: prometheus_global values

sugyan:
Although the value of `prometheus_global` says “It is compatible with the official configuration.”
It seems to be at variance with the default value described at https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file.


gardar:
Any thoughts on this one @SuperQ ?

PR #528, opened 2025-02-07
fix: identation in service file

tbauriedel:
Small PR to fix identation of argument in prometheus.service file

Before:
```
...
  --storage.tsdb.retention.size=0 \
    --web.config.file=/etc/prometheus/web_config.yml \
  --web.console.libraries=/etc/prometheus/console_libraries \
...
```

After:
```
...
  --storage.tsdb.retention.size=0 \
  --web.config.file=/etc/prometheus/web_config.yml \
  --web.console.libraries=/etc/prometheus/console_libraries \
...
```


gardar (approved):
Thanks!

PR #530, opened 2025-02-08
fix: add `enabled: true` to alertmanager

sugyan:
Only in the `alertmanager` was there no written setting that would be enabled on boot.

```bash
$ for f in roles/*/tasks/main.yml; do grep -q 'enabled: true' $f || echo $f; done
roles/alertmanager/tasks/main.yml
```

### DIFF
--- a/roles/alertmanager/tasks/main.yml
+++ b/roles/alertmanager/tasks/main.yml
@@ -55,7 +55,7 @@
     name: alertmanager
     enabled: true
     state: started
-  when:
-    - not ansible_check_mode
   tags:
+    - alertmanager
+    - run
     - alertmanager_run
--- a/roles/alertmanager/tasks/main.yml
+++ b/roles/alertmanager/tasks/main.yml
@@ -47,3 +47,15 @@
     - alertmanager
     - configure
     - alertmanager_configure
+
+- name: Ensure alertmanager is enabled on boot
+  become: true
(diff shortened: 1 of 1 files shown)


roles/alertmanager/tasks/main.yml
```diff
@@ -47,3 +47,15 @@
...
+
+- name: Ensure alertmanager is enabled on boot
+  become: true
+  ansible.builtin.systemd:
+    daemon_reload: true
+    name: alertmanager
+    enabled: true
+    state: started
+  when:
+    - not ansible_check_mode
```
gardar, line 59: The systemd module has full support for check_mode so I don't think this is needed
```suggestion
```

roles/alertmanager/tasks/main.yml
```diff
@@ -47,3 +47,15 @@
...
+  tags:
+    - alertmanager_run
```
gardar, line 61:
```suggestion
  tags:
    - alertmanager
    - run
    - alertmanager_run
```

gardar (changes requested):
Nice catch, thanks!

roles/alertmanager/tasks/main.yml
sugyan: Thanks for the review. I've modified and pushed the part you suggested.

gardar (approved):

PR #548, opened 2025-03-07
feat(process_exporter): add support for recheck-with-time-limit

Frazew:
This CLI flag is useful to prevent process_exporter from putting processes in the wrong groups if it happens to scan them for the first time before they change their command line. This can happen for processes which call exec(3) for example.

The default setting is 0, which leaves the feature disabled. Note that setting a non-zero value for `--recheck-with-time-limit` [automatically also sets `--recheck`](https://github.com/ncabatoff/process-exporter/blob/626431b9a759d425bbb78eb15153f892970aadee/cmd/process-exporter/main.go#L238) so there's no need to set the two CLI flags.


frazew:
Well, the tests now fail because 0.8.0 requires a glibc version too high for debian 11 / almalinux 9... I guess we should therefore keep 0.7.10 in the test, which should now succeed because the setting will be skipped thanks to the added version check, what do you think?

SuperQ:
There should be no glibc requirement at all. Looks like that was a bug in v0.8.0. The next bugfix seems to fix it:

```
$ ldd process-exporter-0.8.1.linux-amd64/process-exporter 
	not a dynamic executable
```

frazew:
oh wow indeed, that wasn't really clear from the changelog, thanks, I've updated the version in the molecule test file

PR #556, opened 2025-03-15
skip_changelog: Update GitHub actions

SuperQ:
Update various github actions to pin SHA for supply chain security.
* Replace `tj-actions/changed-files` due to CVE-2023-51664.

### DIFF
--- a/.github/workflows/conventional-label.yml
+++ b/.github/workflows/conventional-label.yml
@@ -14,7 +14,7 @@ jobs:
       pull-requests: write
     steps:
       - name: "Confirm correct pull request title"
-        uses: mmubeen/action-pr-title@master  # until PR gets merged https://github.com/deepakputhraya/action-pr-title/pull/29
+        uses: deepakputhraya/action-pr-title@077bddd7bdabd0d2b1b25ed0754c7e62e184d7ee
         with:
           allowed_prefixes: "breaking,chore,feat,feature,fix,major,minor,enhancement,\
            deprecated,removed,security,bug,bugfix,docs,packaging,\
@@ -22,7 +22,7 @@ jobs:
 
       - name: "Apply label"
         if: github.event.pull_request.labels.length == 0
-        uses: bcoe/conventional-release-labels@v1
+        uses: bcoe/conventional-release-labels@886f696738527c7be444262c327c89436dfb95a8  # v1.3.1
         with:
           type_labels: |
               {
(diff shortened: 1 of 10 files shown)


gardar (changes requested):
The pinning is great but there's no need to enable dependabot as we already have renovate covering this.

dgl:
> The pinning is great but there's no need to enable dependabot as we already have renovate covering this.

Just FYI; in this incident renovate was partly involved, renovate's default behaviour is to resolve version tags (from comments, see [this](https://docs.renovatebot.com/modules/manager/github-actions/)) and the attacker updated the tag, so renovate made PRs to update the version dep (there's more in depth discussion on the [HN thread](https://news.ycombinator.com/item?id=43367987)). It looks like we don't auto-merge renovate PRs in this repo, but my point is even with renovate the reviewer has to be somewhat aware when reviewing PRs.

.github/workflows/conventional-label.yml
```diff
@@ -54,21 +54,14 @@ jobs:
...
         id: changed-roles
-        uses: tj-actions/changed-files@v45
-        with:
-          path: "roles"
-          diff_relative: "true"
-          files: "**"
-          dir_names: "true"
-          dir_names_max_depth: "1"
-          sha: ${{ github.event.pull_request.head.sha }}
+        run: git diff --name-only ${{ github.event.pull_request.head.sha }} -r | grep -E '^roles/' | cut -f 1-2 -d'/' | sort -u >> "$GITHUB_OUTPUT"
```
dgl, line 61: I haven't tested this, but might need to update actions/checkout to include `fetch-depth` as it does a shallow checkout by default.

electron0zero: I tested this locally and it seems to do the job.

my test setup:
- shallow cloned this repo with --depth=1
- created a branch with name test_123
- modified `roles/blackbox_exporter/handlers/main.yml`  and `roles/blackbox_exporter/tasks/main.yml`
- made commit on my branch with these changes

here is what my git log looked like:
```bash
commit 67cf36f6fd8ea9566420f4f8eff9df959ef69d32
Author: Suraj Nath <9503187+electron0zero@users.noreply.github.com>
Date:   Sun Mar 16 21:05:21 2025 +0530

    test commit

commit 8bb8ea46a2b020fe9db5d7e42a371fdf627fe5ca
Author: gardar <455556+gardar@users.noreply.github.com>
Date:   Wed Mar 5 22:50:40 2025 +0000

    chore: update version

```

here `8bb8ea46a2b020fe9db5d7e42a371fdf627fe5ca` HEAD of main.

```bash
➜  ansible git:(test_123) git diff --name-only 8bb8ea46a2b020fe9db5d7e42a371fdf627fe5ca -r | grep -E '^roles/' | cut -f 1-2 -d'/' | sort -u >> out.t
est
➜  ansible git:(test_123) ✗ cat out.test
roles/blackbox_exporter
```

after this I made one more commit modifying files in `roles/memcached_exporter` and it corrected populated output with:

```bash
roles/blackbox_exporter
roles/memcached_exporter
```

so as far as I understand bash and GitHub actions, this should work fine 🤞🏼

electron0zero (approved):
lgtm, thanks for taking this on.

gardar:
> > The pinning is great but there's no need to enable dependabot as we already have renovate covering this.
> 
> Just FYI; in this incident renovate was partly involved (in addition to the fact the attacker spoofed it), renovate's default behaviour is to resolve version tags (from comments, see [this](https://docs.renovatebot.com/modules/manager/github-actions/)) and the attacker updated the tag, so renovate made PRs to update the version dep (there's more in depth discussion on the [HN thread](https://news.ycombinator.com/item?id=43367987)). It looks like we don't auto-merge renovate PRs in this repo, but my point is even with renovate the reviewer has to be somewhat aware when reviewing PRs.

As I understand it then it updates the tag if you specify the version as a tag in the GitHub actions config, but if you use the sha as a version then it uses that.
So switching out the tags for sha as is being done in this PR should be sufficient.

gardar (approved):

SuperQ:
One review is fine, but it's not really any different as we're the only ones clicking the merge button. If we had auto-merge, adding a review for some things might be nice.

But either way, this CVE issue would not have been prevented by any review here. The supply chain issue was a merged malicious commit. The tag vs sha issue is to prevent tag manipulation. Not to prevent malicious commits.

PR #267, opened 2023-12-19
Allow using web config file

modship:
This PR allow using web config file, in my case i need to use basic auth

Exemple : 
```yml
alertmanager_web_config:
  basic_auth_users:
    John: $2a$12$9kqglntQUzvOFu2cvEYZ0.H6sXuho5Xdc.ID.C2hNtlCYF6q/RXWG
    Doe: $2a$12$sIB6/w8/4Fn5EWaAdrpSGuRqIhkrXzSnQSxOqbiSPhTbdHWkC.oKG
```

PR #576, opened 2025-04-10
Added web configuration for alert manager role

BountyHunter1999:
This PR adds web configuration for the alert manager role, which was missing even though the alert manager has that [configuration](https://prometheus.io/docs/alerting/latest/configuration/#http_config).

This seemed to have been raised before on this [issue](https://github.com/prometheus-community/ansible/pull/267)

This just creates the `web_config.yml` for alertmanager like it does for prometheus


sanchiezes:
Hi all!
How setup web.config.file and cluster.tls-config without templates and variables? Only set path to custom template in extra variables?

rds13:
👍

PR #586, opened 2025-04-26
Set root as default user and group for smartctl exporter

### LINKED ISSUE #298: smartctl exporter role: the default user and group should be root
Just tried out the smartctl exporter role without further configuration and it didn't work. It works after setting `smartctl_exporter_system_user` and ``smartctl_exporter_system_group` to root.

From https://github.com/prometheus-community/smartctl_exporter/issues/146, the upstream package is intended to be run as root, and the provided SystemD unit files in the upstream package are configured to have root as the user as well. There is some extra steps required in order to allow non-root users to run the exporter.

Therefore, I would suggest changing the defaults to `smartctl_exporter_system_user` and ``smartctl_exporter_system_group` to root for a better out-of-the-box experience

n-asenov:
Fixes #298


nasenov:
@gardar @SuperQ could you review the PR?

nasenov:
Closing in favour of #711

Issue #609, opened 2025-06-26
SNMP Exporter: Invalid service up checking during initial diff


Jorich:
I have included prometheus.prometheus.snmp_exporter role in a playbook. 
In CI/CD before execution of playbook a diff is run in following manner: 

`ansible-playbook -i inventory/inventory.yml  playbook-monitoring.yml --check --diff ` 

If that task is run successfully, diff should be reviewed an if all changes are as expected actual playbook could be run. 

`ansible-playbook -i inventory/inventory.yml --vault-id vault/vault-client playbook-monitoring.yml --diff`

With that role, during the check command I get following error: 

`TASK [prometheus.prometheus.snmp_exporter : Ensure snmp_exporter service is started and enabled] ***
fatal: [baa-monitoring]: FAILED! => {"changed": false, "msg": "Could not find the requested service snmp_exporter: host"}`

That is because the check doesn't acutally run the service, hence the error is completely valid. 

Proposition: not run the task "Ensure snmp_exporter service is started and enabled" when running in check mode as it is done similarly in node_exporter role similar task.

Jorich:
Added a fix PR: https://github.com/prometheus-community/ansible/pull/610

gardar:
#610

PR #610, opened 2025-06-26
fix(snmp_exporter): Added when clause to snmp_exporter role

Jorich:
Added when clause to Ensure snmp_exporter service is started and enabled to not run in check mode
https://github.com/prometheus-community/ansible/issues/609


Jorich:
Any ETA when maintainer will be able to approve/decline this PR?

PR #614, opened 2025-06-30
minor: Bump for release

SuperQ:
Do a trivial change tagged minor in order to trigger 0.27.0 release.
* Fixup changelog.yaml

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,14 +4,6 @@ Prometheus.Prometheus Release Notes
 
 .. contents:: Topics
 
-v0.27.0
-=======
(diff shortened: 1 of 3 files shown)


gardar (approved):

gardar:
Hmm I wonder what went wrong as it looks like the 0.26.3 labeling was only present in the github release tag.
The attached collection to that release has the 0.27.0 tag in the filename and contains a changelog and manifest with the 0.27.0 tag and it was also tagged as 0.27.0 in galaxy.

Perhaps a more appropriate cause of action would have been to rename the github release tag?

PR #616, opened 2025-07-09
fix(ipmi_exporter): Fix privilege escalation when running in local mode

### LINKED ISSUE #526: Not possible to run ipmi-exporter as non-root due to NoNewPrivileges in systemd service
I'm trying to make ipmi-exporter run as a non root user (the default if no variables are set by the user).

As per https://github.com/prometheus-community/ipmi_exporter/blob/master/docs/privileges.md, my playbook is like this:

```
- name: Configure IPMI exporter
  hosts: all
  tasks:
    - name: Create sudoers file
      ansible.builtin.copy:
        src: ipmi_exporter.sudoers
        dest: /etc/sudoers.d/ipmi_exporter
        mode: '0400'

- name: Install monitoring stack
  hosts: all
  roles:
    - prometheus.prometheus.node_exporter
    - prometheus.prometheus.ipmi_exporter
  vars:
    ipmi_exporter_modules:
      default:
        collectors:
          - bmc
          - ipmi
          - dcmi
          - chassis
          - sel
        collector_cmd:
          bmc: /usr/bin/sudo
          ipmi: /usr/bin/sudo
          dcmi: /usr/bin/sudo
          chassis: /usr/bin/sudo
          sel: /usr/bin/sudo
```

ipmi_exporter.sudoers:
```
ipmi-exp ALL = NOPASSWD: /usr/sbin/ipmimonitoring,\
                         /usr/sbin/ipmi-sensors,\
                         /usr/sbin/ipmi-dcmi,\
                         /usr/sbin/ipmi-raw,\
                         /usr/sbin/bmc-info,\
                         /usr/sbin/ipmi-chassis,\
                         /usr/sbin/ipmi-sel
```

When applied, I got the following on ipmi_exporter log:
```
Jan 29 16:45:35 [REDACTED] ipmi_exporter[2861019]: ts=2025-01-29T15:45:35.475Z caller=collector_sel.go:60 level=error msg="Failed to collect SEL data" target=[local] error="error running /usr/bin/sudo: exit status 1: sudo: The \"no new privileges\" flag is set, which prevents sudo from running as root.\nsudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.\n"
```

This is due to protections in place on https://github.com/prometheus-community/ansible/blob/main/roles/ipmi_exporter/templates/ipmi_exporter.service.j2

On my tests, 3 lines in systemd service needs to be removed to allow sudo to run:
```
NoNewPrivileges=yes
ProtectKernelModules=true
ProtectKernelTunables=yes
```

I can send a PR, just would like to know what you'd prefer:
- Ansible role default is using ipmi-exp user/group. Therefore, the default should be also to add sudoers.d file in place and remove those 3 lines from systemd
- Another approach is to add a configuration

netcho:
Fixes #526

### DIFF
--- a/roles/_common/tasks/selinux.yml
+++ b/roles/_common/tasks/selinux.yml
@@ -24,6 +24,7 @@
     - "{{ ansible_parent_role_names | first | regex_replace(ansible_collection_name ~ '.', '') }}"
     - configure
     - "{{ ansible_parent_role_names | first | regex_replace(ansible_collection_name ~ '.', '') }}_configure"
+    - selinux
 
 - name: Install selinux python packages [clearlinux]
   ansible.builtin.package:
@@ -40,6 +41,7 @@
     - "{{ ansible_parent_role_names | first | regex_replace(ansible_collection_name ~ '.', '') }}"
     - configure
     - "{{ ansible_parent_role_names | first | regex_replace(ansible_collection_name ~ '.', '') }}_configure"
+    - selinux
 
 - name: Allow port in SELinux
   community.general.seport:
@@ -56,3 +58,4 @@
     - "{{ ansible_parent_role_names | first | regex_replace(ansible_collection_name ~ '.', '') }}"
     - configure
     - "{{ ansible_parent_role_names | first | regex_replace(ansible_collection_name ~ '.', '') }}_configure"
+    - selinux
--- a/roles/ipmi_exporter/defaults/main.yml
+++ b/roles/ipmi_exporter/defaults/main.yml
@@ -13,6 +13,15 @@ ipmi_exporter_modules:
       - chassis
       - sel
 
+ipmi_exporter_sudo_commands:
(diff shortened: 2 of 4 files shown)


SkytAsul:
Would love to see this merged

gardar (changes requested):

gardar:
The sudo policy should not be used for every installation of ipmi_exporter since it only applies to the local mode.
Also, perhaps a better solution would be to just use root, since the [sudo trickery won't work](https://github.com/prometheus-community/ipmi_exporter/blob/291d0107bdab44df5a7ff50ce552ee3ccc23b52a/docs/native.md#privileges) when using the Go-native IPMI implementation.

netcho:
Closing as of now, no intention to work on. If someone wants to solve it, feel free to base your changes on mine and push them forward.

PR #621, opened 2025-07-31
roles/_common: Fix non-boolean conditionals

solj:
This modifies the _common role to use a working conditional for newer ansible versions. Without the fix, you receive the following error:

```
Conditional result was '[...]' of type 'str', which evaluates to True. Conditionals must have a boolean result.
```


gardar:
#727 #764 #769

PR #643, opened 2025-08-26
fix: Conditionals must have a boolean result

PRESIDENT810:
See https://github.com/ansible/ansible/blob/stable-2.19/changelogs/CHANGELOG-v2.19.rst#breaking-changes--porting-guide:

> conditionals - Conditional expressions that result in non-boolean values are now an error by default. Such results often indicate unintentional use of templates where they are not supported, resulting in a conditional that is always true. When this option is enabled, conditional expressions which are a literal None or empty string will evaluate as true, for backwards compatibility. The error can be temporarily changed to a deprecation warning by enabling the ALLOW_BROKEN_CONDITIONALS config option.

This PR follows https://ansible.readthedocs.io/projects/ansible-core/devel/porting_guides/porting_guide_core_2.19.html#broken-conditionals and changed implicit boolean conversions to explicit ones.

### DIFF
--- a/.config/molecule/alternative/prepare.yml
+++ b/.config/molecule/alternative/prepare.yml
@@ -14,7 +14,7 @@
         path: "{{ __cache_path }}"
         state: directory
         mode: 0755
-      when: (__cache_path)
+      when: __cache_path is defined and __cache_path | length > 0
 
     - name: "Fetch binary"
       become: false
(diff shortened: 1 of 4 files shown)


SuperQ:
Fixed in https://github.com/prometheus-community/ansible/pull/619

PR #678, opened 2025-11-07
feat(snmp-exporter): allow skipping snmp config copying step

pabloaul:
This is needed for making it possible to externally manage the snmp config file without it being overwritten by this role

### DIFF
--- a/roles/snmp_exporter/defaults/main.yml
+++ b/roles/snmp_exporter/defaults/main.yml
@@ -8,6 +8,7 @@ snmp_exporter_log_level: info
 
 # If this is empty, role will download snmp.yml file from https://github.com/prometheus/snmp_exporter.
 snmp_exporter_config_file: ""
+snmp_exporter_configure: true
 
 snmp_exporter_binary_install_dir: "/usr/local/bin"
 snmp_exporter_config_dir: "/etc/snmp_exporter"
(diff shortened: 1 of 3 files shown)


SuperQ:
This doesn't really make any sense. Without having ansible manage the config, you can't trigger reloads.

gardar (commented):
It's already possible to skip the configuration using the tags `configure` or `snmp_exporter_configure`

pabloaul:
> This doesn't really make any sense. Without having ansible manage the config, you can't trigger reloads.

Yes, that's the point. Another config generating/handling task would call the handle to restart the snmp-exporter service. The role does not offer a way to use an existing config file from the remote which forces you to have that file on the ansible controller/development machines.




> It's already possible to skip the configuration using the tags `configure` or `snmp_exporter_configure`

My impression was that skipping the snmp configure tag would only be possible using cli args which I feel wouldn't be acceptable to force everyone to specify whenever deploying this to any machine that uses this role. 
Also this would skip the SELinux task which one might prefer to keep.

I am using `ansible.builtin.include_role` however the only ways I found to interact with this role using that was to specify tags to _all_ roles or do parametrization via the variables.

Pretty likely that I am missing something but I unfortunately did not get much wiser from the Ansible docs.

Issue #645, opened 2025-09-02
Role is not idempotent in CI


robpou:
The collection's roles (so _common/install.yml in general) use a cache directory for the downloads and download the binaries to this directory on the controller node, before propagating them to the inventory hosts.

This means when running in a CI pipeline, the plays for creating the local cache path and downloading/unpacking the binary will always count as changes, hence the role is not idempotent.

Am I overlooking something? PR #499 could help by at least letting the inventory hosts perform the problematic steps, but it's been abandoned and that also wouldn't work with `/tmp`, but sure one could take a different local cache path e.g. `/opt` etc.

Could something like

`changed_when: "{{ (lookup('env', 'CI', default='false') | bool) | ternary(false, omit) }}"`

help for these tasks or nah?

gardar:
You should probably set your cache directory to a persistent volume to avoid that...

PR #683, opened 2025-11-18
feature: Add Remote Download for _common and node_exporter

c1rc0le:
Following [PR-499](https://github.com/prometheus-community/ansible/pull/499) which has not been merged due to some conflict.

Currently, the node_exporter installation process always downloads the package to the localhost, unpacks it there, and then uploads it to the remote host.

This change introduces an optional behavior to _common/install, enabling direct download to the remote host. It's now wired for use by node_exporter through the node_exporter_localhost_download variable. The default behavior for node_exporter and other binaries remains unchanged: delegate to localhost. Additionally, these updates allow for customizing the remote cache directory where the binary can be downloaded.

Issues linked :
- [389](https://github.com/prometheus-community/ansible/issues/389)
- [645](https://github.com/prometheus-community/ansible/issues/645)

### DIFF
--- a/roles/_common/meta/argument_specs.yml
+++ b/roles/_common/meta/argument_specs.yml
@@ -69,6 +69,12 @@ argument_specs:
       _common_system_group:
         description: "User group for the system user."
         default: ""
+      _common_remote_download:
+        description: Determines if the package archive is downloaded and extracted on local (false) or remote host (true)"
+        default: "false"
+      _common_remote_cache_path:
+        description: "Remote path to stash the archive and its extraction"
+        default: ""
   preflight:
     short_description: "Internal only - common preflight tasks"
     description: "Internal only - selinux requirements"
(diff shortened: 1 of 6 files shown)


c1rc0le:
Thanks @gardar for your comments. 
I will remove `_common_remote_cache_path `to only keep `_common_local_cache_path`. 

Noted for the bastion option, will change the workflow with : `_common_remote_download_host`.

c1rc0le:
I have reviewed the implementation of the bastion and noticed that it is somewhat complex, as it is not seamlessly integrated within Ansible's unarchive or copy modules. While there may be workarounds using synchronize, these might be error-prone, particularly if the bastion does not establish seamless SSH connections with the target server.

@gardar, would you agree to proceed by implementing `node_exporter_download_host `with localhost and the target server initially, and then explore the possibility of adding the bastion configuration as a second step?

```
node_exporter_download_host: localhost  # (default)
node_exporter_download_host: "{{ inventory_hostname }}"  # download directly on the target

```

PR #707, opened 2026-01-08
feat(blackbox_exporter): add configurable service name

candrii:
Adds blackbox_exporter_service_name variable to allow customizing the systemd service name. Defaults to 'blackbox_exporter' for backwards compatibility.

### DIFF
--- a/roles/blackbox_exporter/defaults/main.yml
+++ b/roles/blackbox_exporter/defaults/main.yml
@@ -9,6 +9,8 @@ blackbox_exporter_web_listen_address: "0.0.0.0:9115"
 blackbox_exporter_system_user: blackbox-exp
 blackbox_exporter_system_group: "{{ blackbox_exporter_system_user }}"
 
+blackbox_exporter_service_name: blackbox_exporter
+
 blackbox_exporter_cli_flags: {}
 # blackbox_exporter_cli_flags:
 #   log.level: "warn"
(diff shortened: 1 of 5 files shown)


gardar:
Thanks for the PR!

This is an intentional design choice. The roles in this collection are deliberately opinionated and we try to keep the number of exposed variables small to avoid unnecessary complexity and maintenance overhead. Customizing the systemd service name is a fairly niche requirement and can be handled outside the role (e.g. with a systemd alias or similar in the playbook).

For that reason we're going to keep the service name fixed in the role and won't add a variable for this.

candrii:
> Thanks for the PR!
> 
> This is an intentional design choice. The roles in this collection are deliberately opinionated and we try to keep the number of exposed variables small to avoid unnecessary complexity and maintenance overhead. Customizing the systemd service name is a fairly niche requirement and can be handled outside the role (e.g. with a systemd alias or similar in the playbook).
> 
> For that reason we're going to keep the service name fixed in the role and won't add a variable for this.

Thanks for the explanation, and I understand the philosophy of keeping variables minimal.

That said, I think the multi-instance use case is worth reconsidering. A systemd alias doesn't fully solve it — if you need to run two instances of the same service on one host with different configurations, you'd still end up with conflicting unit files, config paths, and potentially ports or sockets. The alias approach only gives you a second name pointing to the same unit, not an independent instance.

A single variable for the service name would keep the scope small while unlocking a real deployment pattern that's otherwise impossible without forking or heavily wrapping the role.

Would you be open to reopening this if I keep the change scoped to just the service name, with sensible defaults so it's fully backward-compatible?

PR #754, opened 2026-03-11
feature: allow choosing download host for all roles using _common

ftwrobert:
Following up on PR-633 and PR-499, add the ability to delegate binary downloads to a remote host to address idempotency issues when the ansible controller is ephemeral.


ftwrobert:
The failure condition of the bind_exporter role (HTTP Error 403: rate limit exceeded.) suggests that a re-run of the test may pass?

SuperQ (approved):
Seems fine to me, needs a quick rebase.

gardar (changes requested):
Thanks for taking this on, please add molecule tests for this so that we can confirm that the feature works as expected.

ftwrobert:
I added a scenario to the node_exporter role.

ftwrobert:
The three failing checks appear to be rate limit issues?

PR #789, opened 2026-04-22
feat(ipmi-exporter): add option to run with udev-based device access

guivin:
Add an opt-in flag ipmi_exporter_run_with_capabilities that deploys a udev rule granting the service group read/write access to `/dev/ipmi*` devices. This allows FreeIPMI tools to access IPMI hardware without sudo, while preserving all systemd security hardening.

### DIFF
--- a/roles/ipmi_exporter/defaults/main.yml
+++ b/roles/ipmi_exporter/defaults/main.yml
@@ -30,3 +30,6 @@ ipmi_exporter_config_dir: /etc/ipmi_exporter
 
 # Local path to stash and extract the archive
 ipmi_exporter_local_cache_path: "/tmp/ipmi_exporter-{{ ansible_facts['system'] | lower }}-{{ _ipmi_exporter_go_ansible_arch }}/{{ ipmi_exporter_version }}"
+
+ipmi_exporter_run_with_capabilities: false
+ipmi_exporter_udev_rule_path: "/etc/udev/rules.d/99-ipmi-exporter.rules"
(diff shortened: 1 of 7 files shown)


roles/ipmi_exporter/defaults/main.yml
```diff
@@ -30,3 +30,6 @@ ipmi_exporter_config_dir: /etc/ipmi_exporter
 
 # Local path to stash and extract the archive
 ipmi_exporter_local_cache_path: "/tmp/ipmi_exporter-{{ ansible_facts['system'] | lower }}-{{ _ipmi_exporter_go_ansible_arch }}/{{ ipmi_exporter_version }}"
+
+ipmi_exporter_run_with_capabilities: false
```
SuperQ: [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) is a specific linux feature, this naming is misleading. We're not using `setcap` here.

How about:

```suggestion
ipmi_exporter_udev_enable: false
```

guivin: Done, renamed to `ipmi_exporter_udev_enable` (see [d6e0e48](https://github.com/prometheus-community/ansible/commit/d6e0e48))

roles/ipmi_exporter/handlers/main.yml
```diff
@@ -8,3 +8,12 @@
     state: restarted
   when:
     - not ansible_check_mode
+
+- name: Reload udev rules
+  listen: "reload udev rules"
+  become: true
+  ansible.builtin.command:
```
SuperQ, line 15: Would a systemd restart of `systemd-udev-trigger.service` be enough for this? It's preferred to avoid `command` since it's not idempotent.

guivin: The `--subsystem-match=ipmi` filter was intentional here to limit the trigger to IPMI devices only. Restarting `systemd-udev-trigger.service` would re-trigger all devices system-wide which is broader. 

Since `changed_when: false` is already set, the idempotency concern is addressed at the Ansible reporting level.

I agree command isn't ideal, but there's no clean module equivalent for a scoped `udevadm trigger`

guivin: @SuperQ any additional feedback needed to move this forward ?

PR #791, opened 2026-04-22
feat(prometheus): support tsdb.retention.percentage

### LINKED ISSUE #790: Support config-file retention configuration for Prometheus >=3.11
In Prometheus 3.11, [percentage-based retention policies](https://github.com/prometheus/prometheus/pull/18080) were introduced, although only configurable through the configuration file.
The PR also mentions that the flags `--storage.tsdb.retention.time` and `--storage.tsdb.retention.size` are deprecated, which was done in [October of 2025](https://github.com/prometheus/prometheus/pull/17026).

In order to support the percentage retention, the Prometheus role needs to manage storage configuration in the configuration file, so it makes sense to also move the retention time and size flags to the same file.

rjhenry:
Configure `storage.tsdb.retention` in the configuration file for versions of Prometheus 3.11.0 and greater, rather than deprecated command line flags. This allows for configuring [percentage-based retention](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tsdb).
This closes #790.

### DIFF
--- a/roles/prometheus/templates/prometheus.service.j2
+++ b/roles/prometheus/templates/prometheus.service.j2
@@ -16,11 +16,15 @@ ExecReload=/bin/kill -HUP $MAINPID
 ExecStart={{ prometheus_binary_install_dir }}/prometheus \
 {% if not prometheus_agent_mode %}
   --storage.tsdb.path={{ prometheus_db_dir }} \
+{# If using Prometheus 3.11 or newer, don't set the TSDB retention flags at #}
+{# all, instead use the config file. #}
+{% if prometheus_version is version('3.11.0', '<') %}
 {% if prometheus_version is version('2.7.0', '>=') %}
   --storage.tsdb.retention.time={{ prometheus_storage_retention }} \
   --storage.tsdb.retention.size={{ prometheus_storage_retention_size }} \
 {% else %}
   --storage.tsdb.retention={{ prometheus_storage_retention }} \
+{% endif %}
 {% endif %}
 {% else %}
 {% if prometheus_version is version('3.0.0', '>=') %}
--- a/roles/prometheus/defaults/main.yml
+++ b/roles/prometheus/defaults/main.yml
@@ -23,6 +23,10 @@ prometheus_storage_retention: "30d"
 # [EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Units
 # supported: KB, MB, GB, TB, PB.
 prometheus_storage_retention_size: "0"
+# Available since Prometheus 3.11.0
+# [EXPERIMENTAL] Maximum percentage of total disk space allowed for block
+# storage. Set to 0 to disable percentage-based retention.
+prometheus_storage_retention_percentage: "0"
 
 # The Agent mode optimizes Prometheus for the remote write use case: https://prometheus.io/blog/2021/11/16/agent/
 prometheus_agent_mode: false
--- a/roles/prometheus/meta/argument_specs.yml
+++ b/roles/prometheus/meta/argument_specs.yml
@@ -55,6 +55,12 @@ argument_specs:
           - "Maximum number of bytes that can be stored for blocks."
           - "Units supported: KB, MB, GB, TB, PB."
         default: "0"
+      prometheus_storage_retention_percentage:
+        description:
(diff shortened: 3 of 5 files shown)


rjhenry:
I can't see any `molecule` tests that test for handling of the `prometheus.yml` config file (other than creation), or mention either `prometheus_storage_retention` or `prometheus_storage_retention_size`. That said, I'm not the most familar with Molecule, so it's entirely possible I'm missing something otherwise obvious!
As I couldn't see any tests to mimic, there aren't any introduced tests for the changes in this PR. I'm happy to add some if required, but would appreciate a pointer or two in the right direction (ie: at existing tests I can extend).

roles/prometheus/templates/prometheus.service.j2
```diff
@@ -16,12 +16,14 @@ ExecReload=/bin/kill -HUP $MAINPID
 ExecStart={{ prometheus_binary_install_dir }}/prometheus \
 {% if not prometheus_agent_mode %}
   --storage.tsdb.path={{ prometheus_db_dir }} \
```
gardar, line 18: Do we need to move this to the config too for >= 3.11.0 setups?

roles/prometheus/templates/prometheus.service.j2
```diff
@@ -16,12 +16,14 @@ ExecReload=/bin/kill -HUP $MAINPID
...
+{% if prometheus_version is version('3.11.0', '<') %}
```
gardar, line 21: This should probably be:

```suggestion
{% if prometheus_version is version('3.11.0', '>=') %}
```

gardar (commented):
> I can't see any `molecule` tests that test for handling of the `prometheus.yml` config file (other than creation), or mention either `prometheus_storage_retention` or `prometheus_storage_retention_size`. That said, I'm not the most familar with Molecule, so it's entirely possible I'm missing something otherwise obvious!
> As I couldn't see any tests to mimic, there aren't any introduced tests for the changes in this PR. I'm happy to add some if required, but would appreciate a pointer or two in the right direction (ie: at existing tests I can extend).

It might be a good idea to add this to the `alternative` molecule scenario, actually we're already overwriting the storage retention defaults there, so you could just add the percentage option and bump the prometheus version.

roles/prometheus/templates/prometheus.service.j2
rjhenry: I don't think so - this results in config flags being set for versions of Prometheus less than 3.11, and the use of the config file for versions greater than or equal to 3.11.

| Version | Config |
| ------- | ------|
| 3.9 | Flags |
| 3.10 | Flags |
| 3.11 | File |
| 3.12 | File |

roles/prometheus/templates/prometheus.service.j2
rjhenry: I don't see any way of setting Prometheus' TSDB path in the [current config file reference](https://prometheus.io/docs/prometheus/3.13/configuration/configuration/), and a quick look through the [Prometheus source](https://github.com/prometheus/prometheus/blob/v3.13.0/cmd/prometheus/main.go#L484) doesn't show a deprecation warning like [tsdb.retention](https://github.com/prometheus/prometheus/blob/v3.13.0/cmd/prometheus/main.go#L502). I also can't see any reference to `tsdb.path` in [the config.go's TSDBConfig struct](https://github.com/prometheus/prometheus/blob/v3.13.0/config/config.go#L1149).

roles/prometheus/templates/prometheus.service.j2
rjhenry: I added a Jinja comment to hopefully explain what's going on a little better; probably should have done that from the beginning.

rjhenry:
Updating the Prometheus version used in the alternative tests is failing following the [removal of --alertmanager.timeout](https://github.com/prometheus/prometheus/pull/14640/) in [Prometheus 3.0](https://github.com/prometheus/prometheus/releases/tag/v3.0.0). I'll take a look and fix.

EDIT: After a bit of reading and some local testing, things look good. I've updated the PR with changes. Hopefully, the next run of GitHub Actions tests will succeed.

EDIT 2: I'm ignoring the failures due to GitHub rate limiting of the runners, as I can't re-run those tests specifically.

PR #793, opened 2026-04-24
chore(preflight): Use `vars` lookups

plamerdimak:
Hello, 

I noticed this while installing the role.

Thanks for your feedback.

---

When installing the node exporter, the following warning show up:

    [DEPRECATION WARNING]: The internal "vars" dictionary is deprecated. This feature will be removed from ansible-core version 2.24.
    Origin: /tmp/.ansible/collections/ansible_collections/prometheus/prometheus/roles/_common/tasks/preflight.yml:14:9
    12   ansible.builtin.assert:
    13     that:
    14       - __common_parent_role_short_name ~ '_skip_install' not in vars

               ^ column 9
    [...]
    Use the `vars` and `varnames` lookups instead.

Indeed, Ansible advises to use vars lookups for these assertion checks.

Use `vars` lookups.

Link: https://github.com/ansible/ansible/pull/85673

### DIFF
--- a/roles/_common/tasks/preflight.yml
+++ b/roles/_common/tasks/preflight.yml
@@ -11,7 +11,7 @@
 - name: "Check for deprecated skip_install variable"
   ansible.builtin.assert:
     that:
-      - __common_parent_role_short_name ~ '_skip_install' not in vars
+      - lookup('vars', __common_parent_role_short_name + '_skip_install') | length == 0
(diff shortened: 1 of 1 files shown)


willbuckner:
I raised #869 since this had been sitting for a while, but I don't mean to step on something you're already working on, so if you want to get the tests passing on this one we can use yours instead of mine

PR #805, opened 2026-05-20
enhancement: Expand the supported Ansible and Ubuntu versions

annie444:
This pull request updates the supported Ansible versions and expands test coverage for newer platforms. The main focus is on adding support for Ansible 2.18, 2.19, and 2.20, and introducing Ubuntu 26.04 as a test platform. Additionally, it adjusts version exclusions for several distributions to ensure broader compatibility.

**Ansible version support:**
* Added Ansible versions 2.18, 2.19, and 2.20 to the list of supported versions in `.config/ansible-lint.yml`.

**Test platform updates:**
* Introduced Ubuntu 26.04 as a test platform in `.config/molecule/config.yml`, including a new Docker image configuration.

**Platform-specific Ansible version exclusions:**
* Updated `exclude_ansible_vers` for `almalinux-9`, `debian-11`, and `ubuntu-22.04` to allow all Ansible versions, and added Ubuntu 26.04 with appropriate exclusions.

**Minor configuration improvements:**
* Changed the file mode value to a string in `.config/molecule/alternative/prepare.yml` for better YAML consistency.


annie444:
The molecule and integration tests should run on this PR. However, I can't add the labels in order to do that (and I feel like it would be annoying if I made a minor change to all the roles just to get a label).

.config/molecule/config.yml
```diff
@@ -66,15 +72,17 @@ provisioner:
     host_vars:
       almalinux-9:
         ansible_become_method: su
-        exclude_ansible_vers: ["2.16", "2.17", "2.18", "2.19", "2.20"]
+        exclude_ansible_vers: []
```
SuperQ, line 75: Why? These were added to address specific Python compatibility issues.

SuperQ (changes requested):
The excluded version list is intentional because of Python version compatibility.

gardar:
#807

PR #811, opened 2026-05-31
feat(snmp_exporter):add support for config.expand-environment-variables flag

Luatan:
Adds support to enable the [config.expand-environment-variables ](https://github.com/prometheus/snmp_exporter#configuration) flag to be able to pass environment variables from proemtheus directly.


gardar (changes requested):
Thanks for the PR.
Please add the new option to the argument specs of the role too.

Luatan:
> Thanks for the PR. Please add the new option to the argument specs of the role too.

Hey Gardar,
sure. Added.

Thanks

PR #813, opened 2026-06-04
feat: new nvme-exporter  role

dobbi84:
deploy the E4-Computer-Engineering/nvme-exporter Prometheus exporter following the collection's standard role layout (mirrors the nvidia_gpu_exporter role)

### DIFF
--- a/roles/nvme_exporter/README.md
+++ b/roles/nvme_exporter/README.md
@@ -1,0 +1,53 @@
+<p><img src="https://www.circonus.com/wp-content/uploads/2015/03/sol-icon-itOps.png" alt="graph logo" title="graph" align="right" height="60" /></p>
+
+# Ansible Role: NVMe exporter
+
+## Description
+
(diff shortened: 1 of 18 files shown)


gardar (commented):
Thanks!

Would installing `nvme-cli` before running the role with Molecule allow us to test the exporter behavior a bit more thoroughly? If so, it might be worth adding that to the Molecule prepare step for the `alternative` scenario.

Would it also make sense to add a preflight check for `nvme-cli`? If it is not installed, we could make the role fail, but I think printing a clear `debug` message may be more appropriate than a hard fail, since the exporter can still start even though scrapes may fail and metrics collection will be limited or unavailable.

Regarding the collector configuration, since the exporter currently only has three collectors and all of them are enabled by default, I think it would be cleaner to expose these as explicit booleans instead of using enabled/disabled collector lists.

For example, something along the lines of:

`nvme_exporter_collector_info: true`
`nvme_exporter_collector_smart: true`
`nvme_exporter_collector_ocp: true`

That would make the supported options more obvious to users and avoid cases where the same collector is accidentally listed in both enabled and disabled collectors.

PR #821, opened 2026-06-17
fix: node_exporter preflight checks ssl key as ansible user

epruesse:
The node_exporter TLS certificate private key file may not be readable by the ansible user. The current preflight fails if this is the case. The PR makes the stat call run as the node_exporter_system_user.

### DIFF
--- a/roles/node_exporter/tasks/preflight.yml
+++ b/roles/node_exporter/tasks/preflight.yml
@@ -41,6 +41,8 @@
       ansible.builtin.stat:
         path: "{{ node_exporter_tls_server_config.key_file }}"
       register: __node_exporter_key_file
+      become: true
+      become_user: "{{ node_exporter_system_user }}"
 
     - name: Assert that TLS key and cert are present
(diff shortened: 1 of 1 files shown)


epruesse:
I'm not sure why the tests are failing. Is this expected, or an issue with the test system?

gardar:
> I'm not sure why the tests are failing. Is this expected, or an issue with the test system?

The tests are failing because the service user does not exist yet. The user is created during the installation step of the role, which runs after the preflight checks. The reason this only shows up in the `alternative` test scenario is that TLS support is enabled there, while the `default` and `latest` scenarios do not hit this check.

That said, I think the general idea is a good one. We should verify that the service user can actually access the certificate and key, and ideally this is something we should check consistently across all roles that support TLS.

Maybe the simplest fix here would be to run the certificate/key `stat` checks as root with `become: true`, rather than as the service user. Then we can check not only that the files exist, which we already do, but also that their ownership and permissions allow the service user to use them.

PR #836, opened 2026-06-29
chore: Improve yaml linting

SuperQ:
Implement current required ansible-lint yamllint config changes.

See: https://docs.ansible.com/projects/lint/rules/yaml/

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -14,15 +14,22 @@ rules:
   commas:
     max-spaces-after: -1
     level: error
+  comments:
+    min-spaces-from-content: 1
+  comments-indentation: false
   empty-lines:
     max: 3
     level: error
   hyphens:
     level: error
   line-length:
-    max: 200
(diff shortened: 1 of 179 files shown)


.yamllint.yml
```diff
@@ -14,15 +14,22 @@ rules:
...
+    min-spaces-from-content: 1
+  comments-indentation: false
   empty-lines:
     max: 3
     level: error
   hyphens:
     level: error
   line-length:
-    max: 200
+    max: 250
```
gardar, line 26: Shouldn't this be 160 if we want to follow the ansible-lint defaults?

roles/blackbox_exporter/vars/main.yml
gardar, line 1: In this file I think the previous formatting was a bit more readable. Do we want to strictly enforce the line-length rule here, or allow exceptions where the multiline version is easier to read?

PR #838, opened 2026-06-30
fix: optionally mask diff output for alertmanager

runelan:
During migration towards [MS Teams v2](https://prometheus.io/docs/alerting/latest/configuration/#msteamsv2_config) we discovered that the webhook URL, that should be treated as a secret, gets printed to log output. This PR adds a simple switch for ansible in [diff mode](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_checkmode.html#enforcing-or-preventing-diff-mode-on-tasks), one we rely on during dry-runs but should still prevent printing sensitive information

### DIFF
--- a/roles/alertmanager/defaults/main.yml
+++ b/roles/alertmanager/defaults/main.yml
@@ -146,4 +146,4 @@ alertmanager_local_cache_path: "/tmp/alertmanager-{{ ansible_facts['system'] | l
 alertmanager_system_user: alertmanager
 alertmanager_system_group: "{{ alertmanager_system_user }}"
 
-alert_manager_mask_diff: false
+alertmanager_mask_diff: false
--- a/roles/alertmanager/defaults/main.yml
+++ b/roles/alertmanager/defaults/main.yml
@@ -145,3 +145,5 @@ alertmanager_local_cache_path: "/tmp/alertmanager-{{ ansible_facts['system'] | l
 
 alertmanager_system_user: alertmanager
 alertmanager_system_group: "{{ alertmanager_system_user }}"
+
+alert_manager_mask_diff: false
--- a/roles/alertmanager/tasks/configure.yml
+++ b/roles/alertmanager/tasks/configure.yml
@@ -56,7 +56,7 @@
     mode: 0644
     validate: "{{ alertmanager_binary_install_dir }}/amtool check-config %s"
   no_log: "{{ false if (lookup('env', 'CI')) or (lookup('env', 'MOLECULE_PROVISIONER_NAME')) else true }}"
-  diff: "{{ false if alert_manager_mask_diff else true }}"
+  diff: "{{ alert_manager_mask_diff | bool | default(false) }}"
   become: true
   notify:
     - restart alertmanager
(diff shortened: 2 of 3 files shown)


roles/alertmanager/defaults/main.yml
```diff
@@ -145,3 +145,5 @@ alertmanager_local_cache_path: "/tmp/alertmanager-{{ ansible_facts['system'] | l
 
 alertmanager_system_user: alertmanager
 alertmanager_system_group: "{{ alertmanager_system_user }}"
+
+alert_manager_mask_diff: false
```
SuperQ:
```suggestion
alertmanager_mask_diff: false
```

roles/alertmanager/tasks/configure.yml
```diff
@@ -56,6 +56,7 @@
     mode: 0644
     validate: "{{ alertmanager_binary_install_dir }}/amtool check-config %s"
   no_log: "{{ false if (lookup('env', 'CI')) or (lookup('env', 'MOLECULE_PROVISIONER_NAME')) else true }}"
+  diff: "{{ alert_manager_mask_diff | bool | default(false) }}"
```
SuperQ:
```suggestion
  diff: "{{ alertmanager_mask_diff | bool | default(false) }}"
```

SuperQ (changes requested):

SuperQ:
This also needs to be added to `roles/alertmanager/meta/argument_specs.yml`.

runelan:
> This also needs to be added to `roles/alertmanager/meta/argument_specs.yml`.

Fixed, thanks for the feedback!

roles/alertmanager/tasks/configure.yml
```diff
@@ -56,6 +56,7 @@
     mode: 0644
     validate: "{{ alertmanager_binary_install_dir }}/amtool check-config %s"
   no_log: "{{ false if (lookup('env', 'CI')) or (lookup('env', 'MOLECULE_PROVISIONER_NAME')) else true }}"
+  diff: "{{ alertmanager_mask_diff | bool | default(false) }}"
```
gardar:
```suggestion
  diff: "{{ alertmanager_mask_diff }}"
```

No need to use the default filter since the variable is by default set to `false` under `defaults/main.yml`

gardar (changes requested):

roles/alertmanager/tasks/configure.yml
```diff
@@ -56,6 +56,7 @@
     mode: 0644
     validate: "{{ alertmanager_binary_install_dir }}/amtool check-config %s"
   no_log: "{{ false if (lookup('env', 'CI')) or (lookup('env', 'MOLECULE_PROVISIONER_NAME')) else true }}"
+  diff: "{{ alertmanager_mask_diff }}"
```
gardar: Just realized the logic is actually inverted. As written, diff: `"{{ alertmanager_mask_diff }}"` means mask_diff: true → diff: true, which shows the diff (and prints the secret), and the default false → diff: false, which silently suppresses diffs for everyone. It needs to be negated:

```suggestion
  diff: "{{ not alertmanager_mask_diff }}"
```

That way `alertmanager_mask_diff: true` hides the diff and the false default preserves the current behavior.

gardar (approved):

SuperQ (approved):

PR #845, opened 2026-07-06
refactor(selinux): replace deprecated fact vars

PendaGTP:
### Summary

Currently, `selinux` tasks use some ansible facts without using `ansible_facts`, this cause issue when using `inject_facts_as_vars = False`, since these variables will not be defined.

This PR address this issue.

### DIFF
--- a/roles/_common/tasks/selinux.yml
+++ b/roles/_common/tasks/selinux.yml
@@ -35,7 +35,7 @@
   delay: 2
   become: true
   when:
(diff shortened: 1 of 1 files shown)


PendaGTP:
FYI CI failures are not related to this change but errors comes from rate limiting and timeouts.

SuperQ:
Yes, tests that touch all roles tend to randomly fail due to the load.

PR #848, opened 2026-07-09
feat: Add extra_args variables to nvidia_gpu_exporter to support the new per process stats features

Emalios:
Since version 1.9.0, Nvidia Gpu Exporter allow users to have [per process statistics of vram usage](https://github.com/utkuozdemir/nvidia_gpu_exporter/commit/43dedb0fdc72705befd7064f54459d434c0c8266).

In order to allow this feature you need to add the argument `--collect.compute-apps` to the launch command of nvidia_gpu_exporter, so I propose to add a variable `extra_args` which we can use to activate this option.

I runned the 3 molecules tests, default, latest and alternative. The alternative one needed to up the nvidia_gpu_exporter version cause the feature is only available since 1.9.0

Please do not hesitate to tell me if I did something wrong or if I forgot to do something.


gardar (changes requested):
Thanks for the PR.

We generally try to avoid adding generic `extra_args` / `extra_opts` variables to the roles, since we want the supported options to be explicit and tested. Otherwise, it becomes difficult to know which configurations the role actually supports, and the role starts providing less value compared to users passing all command-line arguments themselves.

That said, I have no issue with supporting this specific option. Please add a dedicated variable for `--collect.compute-apps` instead. Since the feature is only available from version 1.9.0, we should probably guard it with a version check as well.

Please also make sure to add the new option to the role's argument specs, and sign the commits so the DCO check passes.

Emalios:
I changed the extra_args var to an unique boolean which enable the option for the exporter. 
I added a guard in the form of a task firstly executed in main.yml, I don't really know if there is a better way to achieve this

gardar:
> I changed the extra_args var to an unique boolean which enable the option for the exporter. I added a guard in the form of a task firstly executed in main.yml, I don't really know if there is a better way to achieve this

That's fine. The other way to do the version check would have been in the template directly but that wouldn't warn users which attempt to use the option on a unsupported version and it would just be ignored silently, so this is actually the preferred way.

Can you please rebase the branch?

Emalios:
Hello, the branch is rebased

PR #871, opened 2026-08-07
fix: Update molecule driver schema

SuperQ:
The `cgroup_parent` key is not allowed in the new molecule-plugins schema check.


willbuckner:
Any way we could merge this one to unblock the other PRs? :)

gardar (approved):

PR #875, opened 2026-08-13
Fix bad default variable names in configurable download host

elelaysh:
Fixes #2

Note: #754 in upstream has very similar changes. Might revert mine if merged in upstream.


elelaysh:
sorry, bad repo.

PR #883, opened 2026-08-24
feat: Add aerospike_exporter role

anviar:
Add new role: aerospike_exporter
(based on mysqld_exporter role)

### DIFF
--- a/roles/_common/tasks/install.yml
+++ b/roles/_common/tasks/install.yml
@@ -120,7 +120,7 @@
         extra_opts: "{{ _common_binary_unarchive_opts | default(omit, true) }}"
       check_mode: false
       throttle: 1  # see prometheus-community/ansible#457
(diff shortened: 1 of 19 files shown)


anviar:
@gardar hello! is it possible to restart failed checks? (HTTP Error 403: rate limit exceeded)

SuperQ:
I took a look over this exporter and I'm not sure we want to include this.
* It's not a well-written exporter, it has a number of best-practice violations.
* It's not part of the Prometheus ecosystem.
* For tools that are not part of prometheus-community, this is very niche and not a popular system.

This seems like something Aerospike should have their own Ansible collection for.

anviar:
> * It's not part of the Prometheus ecosystem.

same as some others exporters in this project

SuperQ:
Yes, and those other roles are for extremely popular exporters / systems / features.

